PR #57, opened 2022-01-17
Show current selected wallet on UI, switch if another wallet is selected

kujtimprenkuSQA:


kujtimprenkuSQA:
Modified the state management of the wallet selector

Changed how we export the state in src/state/State.ts

Created two methods getState() which gets the current state and updateState() which updates the state.

Added a global function on window object window.updateWalletSelector as a way to notify the Modal.tsx when the state has been changed, and then in the Modal.tsx we can use the current state in the React realm.

Modified all the old references of State module and replaced them with the getState() method.

GentBinakuSQA (approved):

GentBinakuSQA (approved):

PR #63, opened 2022-01-21
Fixed UI styles for dark and light mode

kujtimprenkuSQA:


src/modal/component/Modal/Modal.styles.ts
```diff
@@ -80,90 +84,90 @@ box-sizing: content-box;
...
-    padding: 0.55em 1.4em;
-    text-align: center;
-    color: inherit;
-    font-family: inherit;
-    transition: background 150ms ease-in-out;
-    line-height: 1.15;
-    cursor: pointer;
- }
+  background: inherit;
+  font-size: 14.2px;
```
lewis-sqa, line 108: What's the reason for moving away from em?

src/modal/component/Modal/Modal.tsx
```diff
@@ -274,8 +266,11 @@ function Modal(): JSX.Element {
                 What is a Wallet?
               </span>
               <div
-                className="info-description"
-                style={walletInfoVisible ? mountedStyle : unmountedStyle}
+                className={
+                  walletInfoVisible
+                    ? `info-description show-explanation`
+                    : `info-description hide-explanation`
+                }
```
lewis-sqa, line 273: We could shorten this to:
```jsx
<div className={`info-description ${walletInfoVisible ? "show" : "hide"}-explanation`}>
```

src/modal/component/Modal/Modal.styles.ts
kujtimprenkuSQA: I moved to "px" from "em" because when we integrated the wallet-selector in the guest book example, the font size was affected and the UI would look different meaning in the guest book integration the font size  for buttons was too large and it did not look nice.

src/modal/component/Modal/Modal.tsx
kujtimprenkuSQA: Thank you, will improve it.

lewis-sqa (approved):

PR #69, opened 2022-01-24
Show wallet explanation in example dapp

kujtimprenkuSQA:
Added walletSelectorUI object to the options when we initalize the wallet-selector.

- Provided a custom `description` and `explanation` so they can be shown in the UI.

### DIFF
--- a/example/src/index.js
+++ b/example/src/index.js
@@ -19,6 +19,12 @@ async function initContract() {
       viewMethods: ["getMessages"],
       changeMethods: ["addMessage"],
     },
+    walletSelectorUI: {
+      description: "Please select a wallet to connect to this dapp:",
+      explanation: `Wallets are used to send, receive, and store digital assets. There are different types of wallets. 
+                  They can be an extension added to your browser, a hardware device plugged into your computer, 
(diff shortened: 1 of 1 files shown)


example/src/index.js
```diff
@@ -19,6 +19,12 @@ async function initContract() {
       viewMethods: ["getMessages"],
       changeMethods: ["addMessage"],
     },
+    walletSelectorUI: {
+      description: "Please select a wallet to connect to this dapp:",
+      explanation: `Wallets are used to send, receive, and store digital assets. There are different types of wallets.
```
lewis-sqa, line 24: This string will technically include the whitespace. It might be best to another way using a combination of Arrays and `.join`:

```ts
{
  explanation: [
    "Wallets are used to send, receive, and store digital assets.",
    "There are different types of wallets. They can be an extension",
    "added to your browser, a hardware device plugged into your",
    "computer, web-based, or as an app on your phone."
  ].join(" ")
}
```
It's actually a common way of getting around lint issues where lines can only be 80 or 120 characters long.

lewis-sqa (approved):

PR #70, opened 2022-01-24
Finished adding account id input in ledger wallet option

amirsaran3:


src/modal/component/Modal/Modal.tsx
```diff
@@ -174,10 +162,10 @@ function Modal(): JSX.Element {
                   }
 
                   try {
-                    const ledgerWalletProvider = state.walletProviders[
-                      "ledgerwallet"
-                    ] as ILedgerWallet;
+                    const ledgerWalletProvider = state.walletProviders["ledgerwallet"] as ILedgerWallet;
                     ledgerWalletProvider.setDerivationPath(derivationPath);
+                    ledgerWalletProvider.setAccountId(accountId);
+                    console.log(accountId);
```
lewis-sqa, line 168: Do we still need this `console.log`?

src/wallets/hardware/LedgerWallet.ts
```diff
@@ -73,6 +66,32 @@ export default class LedgerWallet
...
 
+  async setAccountId(accountId: string) {
+    this.accountId = accountId;
+  }
+
+  async checkAccountId(accountId: string, publicKey: string) {
+    const state = getState();
+    const provider = new providers.JsonRpcProvider(`https://rpc.${state.options.networkId}.near.org`);
+
+    const publicKeyString = publicKey;
```
lewis-sqa, line 77: Don't think we need this variable. We can just use `publicKey` directly in the `.query()` parameter object

src/wallets/hardware/LedgerWallet.ts
```diff
@@ -73,6 +66,32 @@ export default class LedgerWallet
...
+        request_type: "view_access_key",
+        finality: "final",
+        account_id: accountId,
+        public_key: publicKeyString,
+      })
+      .catch((err) => {
+        console.log(err);
+      });
+
+    this.nonce = response.nonce;
```
lewis-sqa, line 90: This will crash if the `.catch` path is triggered as `response` will be `undefined`

lewis-sqa (commented):
Just a few minor comments from me.

One general observation is we need to move away from using `any` so much. It seems to be quite common around network calls. Instead, we should start defining the responses so that TypeScript can help us in cases where properties might be null etc. My comment on the crash isn't flagged by TypeScript because `any` means more than "I don't know what the type is". It actually tells TypeScript not to bother checking anything around that variable so you're essentially on your own with the implications that may have

src/modal/component/Modal/Modal.tsx
GentBinakuSQA: I think we should remove all console.log from the code base?

GentBinakuSQA (approved):

GentBinakuSQA (commented):
Besides Lewis comments I have none

GentBinakuSQA (commented):
Besides Lewis Changes I see no issues

lewis-sqa:
@amirsaran3 are you able to address the comments above so we can get this and #71 merged today? We should try to reduce long standing PRs so they don't attract Git conflicts

src/modal/component/Modal/Modal.tsx
amirsaran3: No, it can be removed.

src/wallets/hardware/LedgerWallet.ts
```diff
@@ -72,6 +65,32 @@ export default class LedgerWallet
...
+        account_id: accountId,
+        public_key: publicKey,
+      })
+      .catch((err) => {
+        console.log(err);
+      });
+
+    if (response) {
+      this.nonce = response.nonce;
+    }
```
lewis-sqa, line 89: This should assign `this.nonce` in the `.then` to avoid potential issues if we returned something other than the implicit `undefined` in the `.catch`:

```
await provider
  .query({
    request_type: "view_access_key",
    finality: "final",
    account_id: accountId,
    public_key: publicKey,
  })
  // We really need to move away from using any, but I'll let this slide
  // for now as it's clearly a project-wide problem right now.
  .then((res: any) => {
    this.nonce = res.nonce;
  })
  .catch((err) => {
    console.log(err);
  });
```

amirsaran3: I pushed a commit with these changes.

lewis-sqa (approved):

PR #71, opened 2022-01-25
Task 65/fix near wallet signing with attached near

amirsaran3:


GentBinakuSQA (approved):

kujtimprenkuSQA (approved):

mikedotexe (approved):

lewis-sqa (changes requested):
Looks like there's very similar changes here to #70. Can we address the comments there first. Should hopefully isolate what exactly this PR has changed in comparison too.

mikedotexe:
> Looks like there's very similar changes here to #70. Can we address the comments there first. Should hopefully isolate what exactly this PR has changed in comparison too.

you're right, that other PR also solves that issue for me regarding needing a redirect to Wallet when attaching some NEAR

lewis-sqa (approved):

PR #75, opened 2022-01-26
Bug fixed when dapp user cancels sign in and navigates back

kujtimprenkuSQA:
This PR includes:

- Bug fix when dapp user clicks the Near Wallet option and then navigates back to the dapp  does not crash anymore.
- Update state properly for Near Wallet on page load.

### DIFF
--- a/src/wallets/browser/NearWallet.ts
+++ b/src/wallets/browser/NearWallet.ts
@@ -2,7 +2,7 @@ import BrowserWallet from "../types/BrowserWallet";
 import INearWallet from "../../interfaces/INearWallet";
 import EventHandler from "../../utils/EventHandler";
 import { WalletConnection, Contract } from "near-api-js";
-import { getState } from "../../state/State";
+import { getState, updateState } from "../../state/State";
 
 export default class NearWallet extends BrowserWallet implements INearWallet {
   private wallet: WalletConnection;
@@ -20,7 +20,7 @@ export default class NearWallet extends BrowserWallet implements INearWallet {
   }
 
   async walletSelected() {
-    this.signIn();
+    await this.signIn();
   }
 
   async init() {
@@ -28,6 +28,13 @@ export default class NearWallet extends BrowserWallet implements INearWallet {
     if (!state.nearConnection) return;
     this.wallet = new WalletConnection(state.nearConnection, "near_app");
     EventHandler.callEventHandler("init");
+    if (!this.wallet.isSignedIn()) {
+      updateState((prevState) => ({
+        ...prevState,
+        signedInWalletId: null,
+        isSignedIn: false,
+      }));
+    }
   }
 
   async signIn() {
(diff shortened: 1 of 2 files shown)


src/wallets/browser/NearWallet.ts
```diff
@@ -20,14 +20,21 @@ export default class NearWallet extends BrowserWallet implements INearWallet {
...
     if (!state.nearConnection) return;
     this.wallet = new WalletConnection(state.nearConnection, "near_app");
     EventHandler.callEventHandler("init");
+    if (!this.wallet.isSignedIn()) {
+      updateState((prevState) => ({
+        ...prevState,
+        signedInWalletId: null,
+        isSignedIn: false,
+      }));
+    }
```
lewis-sqa, line 37: I'm concerned we're trying to treat the symptoms here instead of fixing it at the source. Had a quick look over the documentation, but it doesn't really give us much on the nuances to `wallet.requestSignIn`. I'm wondering if we should modify the `signIn` method (which actually isn't really async as the Promise isn't returned) to look something like:

```ts
const state = getState();
this.wallet.requestSignIn(state.options.contract.address).then(() => {
  // My thinking is this block should trigger when the user
  // cancels since they didn't actually sign in.
  if (!this.wallet.isSignedIn()) { 
    return;
  }

  this.setWalletAsSignedIn();
  EventHandler.callEventHandler("signIn");
});
```

src/wallets/browser/NearWallet.ts
```diff
@@ -49,7 +56,7 @@ export default class NearWallet extends BrowserWallet implements INearWallet {
   }
 
   async getAccount(): Promise<any> {
-    if (!this.isConnected()) return null;
+    if (!(await this.isConnected())) return null;
```
lewis-sqa, line 59: This is another case where the method isn't really async. [`this.wallet.isSignedIn()`](https://near.github.io/near-api-js/classes/walletaccount.walletconnection.html#issignedin) actually returns a boolean. We need to remove the `async` keyword on methods that are synchronous (e.g. `isConnected`, `signIn`, `disconnect`, etc).

kujtimprenkuSQA: The reason we are using async in some of the methods NearWallet.ts even though we might not need it is because we are extending BrowserWallet which implements BaseWallet, the async in those methods I think is needed in other wallets like, Ledger or Sender Wallet.

src/wallets/browser/NearWallet.ts
kujtimprenkuSQA: I have followed the suggestion to add the check inside the `then` block of the `requestSignIn` method, then there was another issue with updating the state which we resolved together in a call.
I will push the changes in a bit, Thanks for the help earlier.

lewis-sqa (approved):

Issue #73, opened 2022-01-25
Branded and responsive styling


mikedotexe:
Let's use the NEAR branding guide here:
https://near.org/brand

To come up with a wallet design that's attractive and responsive.

Here's an example from https://1inch.exchange

![image](https://user-images.githubusercontent.com/1042667/151069346-b5993e20-64b5-4459-8e24-67662de0f9f9.png)

PR #90, opened 2022-01-28
Branded and responsive styling

kujtimprenkuSQA:
This PR inclides:

- Changed font-family to Manrope same as https://near.org/brand/
- Changed text, background, border, box-shadow colors with the NEAR palette I got from the above link.
- Changed paddings, margins, leter-spacing, font size, font-weight following the Typograpy section.

These are the only aspects I worked until now in this #73 issue.


GentBinakuSQA (dismissed):

lewis-sqa:
Does this correctly handle styling for `"light"` and `"dark"` themes? I ask because the task description shows us styling for what only looks like a `"dark"` theme

lewis-sqa (approved):

PR #92, opened 2022-01-31
Task 91/event emitter

GentBinakuSQA:


src/utils/__tests__/EventEmitter.unit.ts
```diff
@@ -0,0 +1,23 @@
+
+import { EventHandler } from '../EventsHandler'
+
+describe("emit", () => {
```
lewis-sqa, line 4: We need to add a few more tests in this file. There's still at least 1 more use case for emit (you can have multiple subscribers) but also the other methods: `on` and `off`

src/index.ts
```diff
@@ -11,6 +11,7 @@ export default async function init(options?: Options) {
   const nearConnection = await connect({
     keyStore,
     ...nearConfig,
+    //@ts-ignore
```
lewis-sqa, line 14: We should remove this.

src/core/NearWalletSelector.tsx
```diff
@@ -71,7 +72,7 @@ export default class NearWalletSelector {
     return this.walletController.getAccount();
   }
 
-  on(event: EventList, callback: () => void) {
-    this.walletController.on(event, callback);
+  on(event: any, callback: () => {}) {
```
lewis-sqa, line 75: We should use the `EventList` string literal type here so that users of the library will get autocompletion

src/wallets/browser/NearWallet.ts
```diff
@@ -24,7 +24,8 @@ class NearWallet extends BrowserWallet implements INearWallet {
     const state = getState();
     if (!state.nearConnection) return;
     this.wallet = new WalletConnection(state.nearConnection, "near_app");
-    EventHandler.callEventHandler("init");
+    console.log("emit")
```
lewis-sqa, line 27: I think we put this in to debug the `init` issue but in the end decided to remove the event entirely as there isn't a use case for it yet

src/wallets/hardware/LedgerWallet.ts
```diff
@@ -149,14 +153,13 @@ export default class LedgerWallet extends HardwareWallet implements ILedgerWalle
 
     this.transport.on("disconnect", (res: any) => {
       console.log(res);
-      EventHandler.callEventHandler("disconnect");
+      this.emitter.emit("disconnect", {});
```
lewis-sqa, line 156: Could we remove the empty object here and for the others? We need to think about what the `data` values should be but for now, we can leave them blank (`undefined`).

src/wallets/injected/SenderWallet.ts
```diff
@@ -97,16 +97,8 @@ class SenderWallet extends InjectedWallet implements ISenderWallet {
...
-        if (res.result !== "success") {
-          throw new Error("Failed to sign out");
-        }
-
-        EventHandler.callEventHandler("disconnect");
-
-        return;
-      });
+    this.emitter.emit("disconnect", {});
+    return window[this.injectedGlobal].signOut();
```
lewis-sqa, line 101: Looks like we've regressed on a merge conflict you had regarding my PR that went into dev this morning

src/controllers/WalletController.ts
```diff
@@ -105,8 +110,8 @@ class WalletController {
     return null;
   }
 
-  on(event: EventList, callback: () => void) {
-    EventHandler.addEventHandler(event, callback);
+  on(event: any, callback: () => {}) {
```
lewis-sqa, line 113: We should use the `EventList` string literal type here so that users of the library will get autocompletion

lewis-sqa (changes requested):

src/index.ts
GentBinakuSQA: On my machine it won't build without this ts-ignore

src/wallets/injected/SenderWallet.ts
lewis-sqa: We had the following code before. You'll have hit a conflict and maybe used all of your version:
```ts
return this.wallet.signOut()
      .then((res) => {
        if (res.result !== "success") {
          throw new Error("Failed to sign out");
        }

        EventHandler.callEventHandler("disconnect");

        return;
      });

```

lewis-sqa (approved):

lewis-sqa (approved):

src/wallets/injected/SenderWallet.ts
```diff
@@ -63,7 +65,6 @@ class SenderWallet extends InjectedWallet implements ISenderWallet {
 
     await this.setWalletAsSignedIn();
 
-    EventHandler.callEventHandler("signIn");
```
lewis-sqa, line 66: Missed this one! Will be resurrected in #104

PR #103, opened 2022-02-03
Logger Service

GentBinakuSQA:


src/utils/Logger.ts
```diff
@@ -0,0 +1,146 @@
+import { EventEmitter } from 'events';
```
lewis-sqa, line 1: I'm not sure we need such a complex logging system in the project just yet. I realise you may have got this code from here: https://adrianhall.github.io/cloud/2019/06/30/building-an-efficient-logger-in-typescript/?

It really just need to be a singleton module that exports a few methods for the different of logs. I mean `log` and `error` would suffice for now. It's just to allow logging in the project while having the ability to turn them off for production and in tests

Issue #100, opened 2022-02-02
Migrate Away From View Method Functionality In Wallets


lewis-sqa:
## Description

View Method functionality is coupled with each wallet integration. This abstraction isn't quite right as view methods don't concern wallets - there's nothing to sign. as view methods directly talk to RPC Endpoints. I propose that we move to a single connection further up in the stack such as `NearWalletSelector` or `SmartContract`.

## Acceptance Criteria

- `.view` is removed from each wallet implementation.
- `SmartContract` is simplified to use the single connection when executing the `.view` method.

lewis-sqa:
Useful Cookbook resource: https://github.com/near/near-api-js/blob/master/examples/cookbook/utils/get-state.js

lewis-sqa:
Considering the idea of an `RpcService` that abstracts the different calls we need and handles parsing the responses:
- `view`: https://docs.near.org/docs/api/rpc/contracts#call-a-contract-function
- `getAccessKey` (Ledger Wallet account id validation): https://docs.near.org/docs/api/rpc/access-keys#view-access-key
- `getAccount`: https://docs.near.org/docs/api/rpc/contracts#view-account

PR #104, opened 2022-02-03
#100 Migrate View Functionality

lewis-sqa:
## This PR Includes

- Added a `ProviderService` class to abstract the various RPC calls required in each wallet and `Contract`. This also provides better type support over the `providers.JsonRpcProvider` class from `near-api-js`.
- Added full support for `getAccount` on each wallet.
- Updated the contract configuration to simply `accountId`.
- Updated the wallets to use of the new `ProviderService`.
- Fixed missing "signIn" event for `SenderWallet`.
- Fixed the `getConfig` call(s) so that they use `options.networkId`.
- Removed a several config aliases.

### DIFF
--- a/src/services/provider/ProviderService.spec.ts
+++ b/src/services/provider/ProviderService.spec.ts
@@ -5,7 +5,10 @@ import ProviderService, {
   ViewAccountParams,
 } from "./ProviderService";
 import { mock } from "jest-mock-extended";
-import { JsonRpcProvider } from "near-api-js/lib/providers";
+import {
+  FinalExecutionOutcome,
+  JsonRpcProvider,
+} from "near-api-js/lib/providers";
 import { providers } from "near-api-js";
 import {
   createQueryResponseMock,
@@ -13,6 +16,11 @@ import {
   createViewAccessKeyResponseMock,
   createViewAccountResponseMock,
 } from "./providerServiceMocks";
+import { SignedTransaction } from "near-api-js/lib/transaction";
+import {
+  BlockReference,
+  BlockResult,
+} from "near-api-js/lib/providers/provider";
 
 const defaults = {
   url: "https://rpc.testnet.near.org",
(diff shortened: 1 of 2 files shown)


src/core/Contract.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  view({ methodName, args }: Omit<ViewParams, "contractId">) {
+    return this.provider.callFunction({
+      accountId: this.accountId,
+      methodName,
+      args
+    });
+  }
+
+  call({ actions }: Omit<CallParams, "receiverId">) {
```
lewis-sqa, line 27: Add `async` keyword so that exceptions become rejections.

src/index.ts
```diff
@@ -4,8 +4,8 @@ import { connect, keyStores } from "near-api-js";
 import { updateState } from "./state/State";
 import Options from "./types/Options";
 
-export default async function init(options?: Options) {
-  const nearConfig = getConfig(process.env.NEAR_ENV || "testnet");
+export default async function init(options: Options) {
+  const nearConfig = getConfig(options.networkId);
```
lewis-sqa, line 8: My next task will be to remove this async init entirely. To avoid fixing it in this PR, I've duplicated the `getConfig` call in here and `NearWalletSelector`.

src/interfaces/IWallet.ts
```diff
@@ -29,7 +28,6 @@ export default interface IWallet {
   isConnected(): Promise<boolean>;
   setWalletAsSignedIn(): Promise<void>;
   signIn(): Promise<void>;
-  getAccount(): Promise<any>;
-  view(params: ViewParams): Promise<any>;
+  getAccount(): Promise<AccountInfo | null>;
```
lewis-sqa, line 31: Another `any` bites the dust!

src/wallets/hardware/LedgerWallet.ts
```diff
@@ -351,21 +327,18 @@ export default class LedgerWallet
...
     });
 
-    const signedSerializedTx = signedTransaction.encode();
+    const res = await this.provider.sendTransaction(signedTransaction);
 
-    const base64Response: any = await provider.sendJsonRpc(
-      "broadcast_tx_commit",
-      [Buffer.from(signedSerializedTx).toString("base64")]
-    );
+    if (typeof res.status !== "string") {
```
lewis-sqa, line 332: So after removing the `any` type here, it seems `status` can be either an object or a string. I need your help here @amirsaran3 to understand what the original code was trying to do as previously a crash was possible when `res.status` is a string for whatever reason (`Buffer.from` expects a string not undefined)

GentBinakuSQA (approved):

PR #112, opened 2022-02-04
Logging

GentBinakuSQA:


src/utils/Logger.ts
```diff
@@ -0,0 +1,30 @@
...
+  public log(msg: string, ...supportingData: any[]): void {
+    this.emitLogMessage("log", msg, supportingData);
+  }
+  private emitLogMessage(
+    msgType: "debug" | "info" | "warn" | "error" | "log",
+    msg: string,
+    supportingData: any[]
+  ) {
+    if (supportingData.length > 0) {
+      console[msgType](msg, supportingData);
```
lewis-sqa, line 25: We should spread `supportingData` here

src/utils/Logger.ts
```diff
@@ -0,0 +1,30 @@
+import { LogInterface } from "../interfaces/LogInterface";
+
+export class Logger implements LogInterface {
+  public debug(msg: string, ...supportingData: any[]): void {
```
lewis-sqa, line 4: We don't need to use the `public` keyword in classes. In JavaScript/TypeScript this is the default behaviour

src/utils/Logger.ts
```diff
@@ -0,0 +1,30 @@
...
+    this.emitLogMessage("debug", msg, supportingData);
+  }
+  public info(msg: string, ...supportingData: any[]): void {
```
lewis-sqa, line 7: We don't need to define the return type as it's implicitly defined which is usually preferred

src/utils/Logger.ts
```diff
@@ -0,0 +1,30 @@
...
+    this.emitLogMessage("info", msg, supportingData);
+  }
+  public warn(msg: string, ...supportingData: any[]): void {
```
lewis-sqa, line 10: We should avoid using `any`. It's considered a warning in ESLint for now as there's existing cases that need addressing

src/utils/Logger.ts
```diff
@@ -0,0 +1,30 @@
...
+    this.emitLogMessage("warn", msg, supportingData);
+  }
+  public error(msg: string, ...supportingData: any[]): void {
+    this.emitLogMessage("error", msg, supportingData);
+  }
+  public log(msg: string, ...supportingData: any[]): void {
+    this.emitLogMessage("log", msg, supportingData);
+  }
+  private emitLogMessage(
+    msgType: "debug" | "info" | "warn" | "error" | "log",
```
lewis-sqa, line 20: We should extract this into a type literal

src/utils/__tests__/Logger.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+  });
+  afterAll(() => {
+    console.log = log; // restore original console.log after all tests
+    console.warn = warn;
+    console.error = error;
+    console.debug = debug;
+    console.info = info;
+  });
+  it("no log", () => {
+    // TODO: test something that should not log
```
lewis-sqa, line 25: Looks like there's still some where to be done in the tests

src/utils/__tests__/Logger.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    expect(console.log).not.toHaveBeenCalled();
+  });
+  it("some log", () => {
+    // TODO: test something that should log
+    logger.log("something");
+    expect(console.log).toHaveBeenCalled();
```
lewis-sqa, line 31: We should use `toHaveBeenCalledWith` so we can assert it's called with the correct parameters

lewis-sqa (changes requested):
It looks like you've copied the solution described in this article: https://blog.hellojs.org/simple-javascript-logger-in-typescript-demonstrating-interfaces-union-types-and-rest-parameters-6efc5aee2c97. We need to stop doing this as these solutions often provide a generalised approach or solve a slightly different problem to what we have.

We're still missing:
- Logic that allows us to turn this off in production.
- Logic that allows us to turn this off during tests (other than the unit tests for it).
- Replace usages of `console` throughout the project.

src/utils/Logger.ts
lewis-sqa: `doActive` is still using this.

src/utils/Logger.ts
lewis-sqa: Still a few methods with this return type.

src/utils/Logger.ts
lewis-sqa: Think we can use `unknown` instead of `any`.

src/utils/__tests__/Logger.spec.ts
lewis-sqa: We should reference the parameters we pass in the assertions

src/utils/Logger.ts
```diff
@@ -0,0 +1,35 @@
...
+  private emitLogMessage(
+    msgType: MessageType,
+    msg: string,
+    supportingData?: any[]
+  ) {
+    if (this.doActive) {
+      if (supportingData) console[msgType](msg, ...supportingData);
+      else console[msgType](msg);
+    } else
+      console[msg] = () => {
```
lewis-sqa, line 31: I don't think we need this else logic at all as we just want logging  to be no-op if it's disabled

rgolea:
Should we also have a `@Log` or `@Debug` decorators to activate if, say for example we set up a `DEBUG=true` as global variables? Would be nice to see all the steps that are triggered throughout the project. @lewis-sqa

src/interfaces/LogInterface.ts
```diff
@@ -0,0 +1,6 @@
+export interface LogInterface {
+  debug(msg: string, supportingData?: any[]): void;
+  warn(msg: string, supportingData?: any[]): void;
```
lewis-sqa, line 3: We'll need to update the `any` usages here to use `unknown`

src/interfaces/LogInterface.ts
```diff
@@ -0,0 +1,6 @@
+export interface LogInterface {
```
lewis-sqa, line 1: We should avoid using `Interface` in types/interface names. Given we don't have multiple implementations of loggers we could actually remove this entirely. `Logger` defines it's own interface and avoids the issue we have now were we have to update types in the implementation interface (`any` -> `unknown` comment)

src/utils/Logger.ts
```diff
@@ -0,0 +1,32 @@
+import { LogInterface } from "../interfaces/LogInterface";
+import MessageType from "../types/MessageType";
+
+export class Logger implements LogInterface {
+  public doActive: boolean;
```
lewis-sqa, line 5: We don't need the `public` modifier here as JavaScript/TypeScript classes are public by default

src/utils/__tests__/Logger.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+  logger.doActive = true;
+
+  beforeEach(() => {
+    console.log = jest.fn(); // create a new mock function for each test
+    console.warn = jest.fn();
+    console.error = jest.fn();
+    console.debug = jest.fn();
+    console.info = jest.fn();
+  });
+  afterAll(() => {
```
lewis-sqa, line 20: Consider switching this to `afterEach` and including the `logger.doActive` assignments in these lifecycle hooks. For example

```ts
beforeEach(() => {
  logger.doActive = true;
});

afterEach(() => {
  logger.doActive = false;
});
```
This way you don't have to worry if a test changes it (no log in production) and forgets to reset for the next test.

Issue #117, opened 2022-02-08
Simplify State Model


lewis-sqa:
## Description

Most of the properties in state aren't actually stateful or at the very least don't need to be global:
- `options` is a parameter to `NearWalletSelector` and should be delegated throughout the library not placed in state to be mutable.
- `walletProviders` is implementation detail of `WalletController`.
- `isSignedIn` and `signedInWalletId` is stateful but isn't the source of truth. Consumers should use the new storage provider. To be handled in another task that includes using what was created in #95 .

The remaining properties are really Modal UI state which should live in React. However this needs more thought as I suspect we need to reshuffle what's stateful in the wallet implementations (e.g. Ledger `accountId` and `derivationPath`) and what's shared between the two realms. These are therefore out of scope in this task and might instead be tackled as part of #54.

## Acceptance Criteria

- `options` is removed from state.
- `walletProviders` is removed from state.

lewis-sqa:
This task has been parked while I work on solving #64. The open PR has caused major regressions with the Ledger wallet too which will need to be fixed before merging.

PR #120, opened 2022-02-08
#117 Simplify State Model

lewis-sqa:


src/wallets/browser/NearWallet.ts
```diff
@@ -56,9 +53,7 @@ class NearWallet extends BrowserWallet implements INearWallet {
   }
 
   async signIn() {
-    const state = getState();
-
-    this.wallet.requestSignIn(state.options.accountId).then(() => {
+    this.wallet.requestSignIn(this.options.accountId).then(() => {
```
kujtimprenkuSQA, line 56: The `requestSignIn` method fails probably we need to call `this.init()` somwhere before trying to access properties of `this.wallet`

kujtimprenkuSQA (approved):

PR #121, opened 2022-02-09
Creating logging service

### LINKED ISSUE #101: Create Logging Service
## Description

Logs can be extremely useful when debugging problems in the project. To continue using this approach, we need to consider moving to a service that provides a minor abstraction around the `console` API. The main purpose of this service is to allow us to turn these logs off in production.

## Acceptance Criteria

- New singleton `Logger` service that supports `console.log`, `console.info`, and `console.warn` `console.error`.
- Replace existing console calls with logger instance.

rgolea:
closes #101 
replaces #112

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -51,7 +51,7 @@ module.exports = {
     "default-case": "off",
     "no-caller": "error",
     "no-case-declarations": "off",
-    "no-console": ["error", { allow: ["error", "info", "warn"] }],
+    "no-console": "error",
     "no-debugger": "error",
     "no-eval": "error",
     "no-fallthrough": "error",
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -51,7 +51,7 @@ module.exports = {
     "default-case": "off",
     "no-caller": "error",
     "no-case-declarations": "off",
-    "no-console": ["warn", { allow: ["error", "info", "warn"] }],
+    "no-console": ["error", { allow: ["error", "info", "warn"] }],
     "no-debugger": "error",
     "no-eval": "error",
     "no-fallthrough": "error",
--- a/src/services/logging.service.spec.ts
+++ b/src/services/logging.service.spec.ts
@@ -16,10 +16,6 @@ describe("LoggingService Unit Tests", () => {
     mockReset(internal);
   });
 
-  it("should be defined", () => {
-    expect(Logger).toBeDefined();
-  });
-
   it("should be a singleton", () => {
     const logger2 = new Logger(internal);
     expect(logger).toEqual(logger2);
--- a/src/services/logging.service.spec.ts
+++ b/src/services/logging.service.spec.ts
@@ -1,4 +1,4 @@
-import { Logger } from "./logging.service";
+import { Logger, logger as defaultLogger } from "./logging.service";
 import { mock, mockReset } from "jest-mock-extended";
 
 describe("LoggingService Unit Tests", () => {
@@ -10,8 +10,11 @@ describe("LoggingService Unit Tests", () => {
     logger = new Logger(internal);
   });
 
-  afterEach(() => {
+  beforeEach(() => {
     Logger.debug = true;
+  });
+
+  afterEach(() => {
     // Reset the mock after the JS stack has cleared
     mockReset(internal);
   });
@@ -50,5 +53,9 @@ describe("LoggingService Unit Tests", () => {
   it("should error", () => {
     logger.error("test");
     expect(internal.error).toHaveBeenCalledWith("test");
+  });
+
+  it("should export a default instance", () => {
+    expect(defaultLogger).toBeInstanceOf(Logger);
   });
 });
--- a/src/services/logging.service.spec.ts
+++ b/src/services/logging.service.spec.ts
@@ -1,65 +1,58 @@
 import { Logger } from "./logging.service";
 import { mock, mockReset } from "jest-mock-extended";
 
-describe('LoggingService Unit Tests', () => {
+describe("LoggingService Unit Tests", () => {
   let logger: Logger;
-  let hiddenLogger: Logger;
   let internal: Console;
 
   beforeAll(() => {
     internal = mock<Console>();
     logger = new Logger(internal);
-    hiddenLogger = new Logger(internal, true);
   });
 
   afterEach(() => {
+    Logger.hidden = false;
     // Reset the mock after the JS stack has cleared
     mockReset(internal);
   });
 
-  it('should be defined', () => {
+  it("should be defined", () => {
     expect(Logger).toBeDefined();
   });
 
-  it('should be a singleton', () => {
+  it("should be a singleton", () => {
     const logger2 = new Logger(internal);
     expect(logger).toEqual(logger2);
   });
 
-  it('should hide logs if hidden', () => {
-    hiddenLogger.log('test');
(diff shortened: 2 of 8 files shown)


src/services/logging.service.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+    hiddenLogger.log('test');
+    expect(internal.log).not.toHaveBeenCalled();
+  });
+
+  it('should show logs if not hidden', () => {
+    logger.log('test');
+    expect(internal.log).toHaveBeenCalled();
+  });
+
+  it('should have a log, info, warn and error method', () => {
```
lewis-sqa, line 39: We can omit this test as TypeScript checks this for us 🙂

lewis-sqa (commented):
Looks good 👍  Few points from me:

- Need to resolve the lint errors, ideally before transitioning from draft to open.
- Avoid using `any` (this will become an error shortly).
- I'm interested to see how each module will be able to easily import this without having to declare if it's hidden each time. Maybe we expose `Logger` as a named export and have the default export predefined to be hidden in production/tests? It's a common way of emulating the singleton pattern without having to implement logic like `instances` (I suppose we haven't really got a singleton here as you can have two instances 😛)

rgolea:
> Looks good 👍 Few points from me:
> 
> * Need to resolve the lint errors, ideally before transitioning from draft to open.

Will do a PR soon with vscode changes

> * Avoid using `any` (this will become an error shortly).

Solved by using `unknown`

> * I'm interested to see how each module will be able to easily import this without having to declare if it's hidden each time. Maybe we expose `Logger` as a named export and have the default export predefined to be hidden in production/tests? It's a common way of emulating the singleton pattern without having to implement logic like `instances` (I suppose we haven't really got a singleton here as you can have two instances 😛)

Yes, so I found that doing `Logger.debug` would be the way to go. As anyone who wants to debug can go and import the service and set `debug` to `true` as it is fairly encapsulated. I removed the 2 instances thing. I think it's a better approach this way but let me know what you think.

src/services/logging.service.ts
```diff
@@ -0,0 +1,40 @@
+export class Logger
+  implements Pick<Console, "log" | "info" | "warn" | "error">
+{
+  private static instance: Logger;
+  private static _hidden = false;
```
lewis-sqa, line 5: I was thinking we'd have a dynamic default based on production vs. development. Testing would be off by default too (the tests for the module would just turn it on). I can't say at this point as there are some lint errors, but I suspect unit tests (if there are logs in the modules under test) will contain logs

src/wallets/browser/NearWallet.ts
```diff
@@ -17,6 +17,9 @@ import {
 } from "../../interfaces/IWallet";
 import ProviderService from "../../services/provider/ProviderService";
 import getConfig from "../../config";
+import { Logger } from "../../services/logging.service";
+
+const logger = new Logger();
```
lewis-sqa, line 22: We could remove this import and new up by using a default export:

```ts
// logging.service.ts
export class Logger { /* ... */ }

export default new Logger();

// ---------------------

// module-that-wants-to-use-logger.ts
import logger from ".../services/logging.service";

logger.log("Hello World!");

// ---------------------

// logging.service.spec.ts
import { Logger } from "./logging.service";

beforeEach(() => {
  logger = new Logger();
  logger.debug = true;
});

afterEach(() => {
  logger.debug = false;
});

// ....
```

src/wallets/injected/SenderWallet.ts
```diff
@@ -29,13 +31,14 @@ class SenderWallet extends InjectedWallet implements ISenderWallet {
...
 
+
     this.onAccountChanged();
 
     this.onNetworkChanged();
 
     return this.wallet
       .init({ contractId: state.options.accountId })
-      .then((res) => console.log("SenderWallet:init", res));
+      .then((res) => logger.log("SenderWallet:init", res));
```
lewis-sqa, line 41: Missing a couple of other console calls in this module

lewis-sqa (commented):
Few more comments from me. We should switch `console` to be an error to avoid missing anymore and get closer to eradicating the lint warnings. Since the module itself will trigger errors, we'll want to use something like to let ESLint know it's an exception to the rule:
```ts
// eslint-disable-line no-console
```

src/services/logging.service.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  afterEach(() => {
+    Logger.debug = true;
+    // Reset the mock after the JS stack has cleared
+    mockReset(internal);
+  });
+
+  it("should be defined", () => {
+    expect(Logger).toBeDefined();
+  });
```
lewis-sqa, line 21: We can remove this test as it's also covered by TypeScript. I guess generally speaking this matcher is rarely needed in a TypeScript project (if at all?). The same applies for testing methods with incorrect parameter types (although tests are written in TypeScript so they won't actually compile 😆)

src/services/logging.service.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+  let logger: Logger;
+  let internal: Console;
+
+  beforeAll(() => {
+    internal = mock<Console>();
+    logger = new Logger(internal);
+  });
+
+  afterEach(() => {
+    Logger.debug = true;
```
lewis-sqa, line 14: This should be `false` (reset back to default). I guess this is the downside of using a truly Singleton pattern unless you get Jest to reset the module. Using the named export + default export approach mentioned in another comment would allow you to new up the `Logger` for each test without side-effects

src/services/logging.service.ts
rgolea: I mean, you can always just turn on/off by assigning `Logger.debug` to `true`. At any given moment. I did not really know exactly where to tap into `process.env` and I believe we'll be using the logger at development so it's easy to just turn on at the start of the file. Not really sure how to go from here.

src/wallets/browser/NearWallet.ts
rgolea: The reason I left it really was to be able to turn on a new logger instance outside the project and use a custom logging method like sentry to get logs from this module. That is the reason I would not instanciate in the same file. You could just import beforehand the Logger and then instanciate the `NearWalletSelector` library.

src/wallets/injected/SenderWallet.ts
rgolea: Had them locally. Forgot to commit them. Thanks!

src/services/logging.service.spec.ts
rgolea: I generally do that one because I create the file at the same time as the code and jest crashes if there is an empty spec file. 👍

src/services/logging.service.spec.ts
rgolea: You are right. I was overcomplicating things. Switching right now.

.eslintrc.js
```diff
@@ -51,7 +51,7 @@ module.exports = {
     "default-case": "off",
     "no-caller": "error",
     "no-case-declarations": "off",
-    "no-console": ["warn", { allow: ["error", "info", "warn"] }],
+    "no-console": ["error", { allow: ["error", "info", "warn"] }],
```
lewis-sqa, line 54: As discussed, this should be:
```
"no-console": "error"
```

lewis-sqa (approved):

PR #122, opened 2022-02-09
Handle account change in Sender Wallet

kujtimprenkuSQA:
## This PR includes

- Listen to the onAccountChanged event emited by Sender Wallet
- If the current selected account is not signed in prompt user to connect dApp with the newly selected account.
- If user rejects fire event "disconnect" to inform dApp
- Fixed typo in example dApp when listening on the `"disconnect"` event.

## Notes

I tried many different ways to handle the case when user rejects (closes the sender wallet popup window) during the process when a new account is selected inside the Sender Wallet the best I could come up was to call the `disconnect`in which case the dApp sets account to null.

### DIFF
--- a/src/wallets/injected/SenderWallet.ts
+++ b/src/wallets/injected/SenderWallet.ts
@@ -69,11 +69,11 @@ class SenderWallet extends InjectedWallet implements ISenderWallet {
 
   async signIn() {
     const state = getState();
-    const { accessKey } = await this.wallet.requestSignIn({
+    const response = await this.wallet.requestSignIn({
       contractId: state.options.accountId,
     });
 
-    if (!accessKey) {
+    if (response.error) {
       if (this.wallet.isSignedIn()) await this.disconnect();
       return;
     }
--- a/src/wallets/injected/SenderWallet.ts
+++ b/src/wallets/injected/SenderWallet.ts
@@ -36,9 +36,7 @@ class SenderWallet extends InjectedWallet implements ISenderWallet {
 
     const state = getState();
 
-    this.wallet.onAccountChanged((newAccountId) => {
-      console.log("SenderWallet:onAccountChange", newAccountId);
-    });
+    this.onAccountChanged();
 
     this.onNetworkChanged();
 
@@ -78,7 +76,6 @@ class SenderWallet extends InjectedWallet implements ISenderWallet {
     }
 
     await this.setWalletAsSignedIn();
-    this.emitter.emit("signIn");
 
     this.emitter.emit("signIn");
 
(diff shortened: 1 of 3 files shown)


src/wallets/injected/SenderWallet.ts
```diff
@@ -114,6 +112,26 @@ class SenderWallet extends InjectedWallet implements ISenderWallet {
...
+        .init({
+          contractId: state.options.accountId,
+        })
+        .then((response) => {
+          if (response.accessKey) {
+            this.emitter.emit("signIn");
+          } else {
+            this.signIn().then();
+          }
+        });
```
lewis-sqa, line 131: I think we could reuse `this.signIn` here?

src/wallets/injected/SenderWallet.ts
```diff
@@ -76,11 +74,11 @@ class SenderWallet extends InjectedWallet implements ISenderWallet {
     });
 
     if (!accessKey) {
+      if (this.wallet.isSignedIn()) await this.disconnect();
```
lewis-sqa, line 77: I think I made a mistake reusing `InitResponse` for `requestSignIn`. I think the idea of getting no `accessKey` back is wrong for this method (only occurs for `init`). I also think disconnecting here might have some weird implications if signing out displays a prompt in the future for SenderWallet. The user would go to login only to be shown a logout flow

lewis-sqa (commented):
Couple of comments from me 🙂

src/wallets/injected/SenderWallet.ts
kujtimprenkuSQA: You're right it can work by calling `this.signIn`  in at this point too. I was trying to figure out what to do  when a user closes the sender wallet window so I left that if else check but I see we're doing the same in `this.signIn`.

src/wallets/injected/SenderWallet.ts
kujtimprenkuSQA: If the user signs in successfully than the `InitResponse `is correct but if the user closes the window without signing in the response we get back is entirely different.
It returns only this:

```javascript
{
    "error": "User reject",
    "notificationId": 1644402625554,
    "type": "sender-wallet-result"
}
```

As for the disconnecting at this point works fine, but I don't know what the Sender Wallet has in store for the future.

lewis-sqa: That's rather annoying. I think we need to take a closer look at the types for this wallet (if only Sender Wallet had documentation). We'll basically need to check if there's an `error` property in the response. If there is, we should throw straight away (or in this case simply return as cancelling isn't really an "error"). I'd suggest creating a separate interface for the method:

```ts
interface AccessKey =  {
  publicKey: {
    data: Uint8Array;
    keyType: number;
  };
  secretKey: string;
}

export interface InitResponse {
  accessKey: AccessKey | "";
}

export type RequestSignInResponse =
  | { accessKey: AccessKey; }
  | {
      error: string;
      notificationId: number;
      type: "sender-wallet-result";
    };
```

lewis-sqa (approved):

PR #123, opened 2022-02-09
Added vscode configurations for the workspace

### LINKED ISSUE #110: Create editor specific settings
## Description

These should be just some small settings that can allow faster and more reliable workflow experience.

## Acceptance criteria

The added configs should only he absolutely necessary settings that can help with the workflow (eg: desctivate the prettier plugin because it is included in ESLint so configurations might be conflicting)

rgolea:
closes #110

Please let me know if you would remove some of them or allow them all. Also, there is still a need for other editors as well.


lewis-sqa:
Added @amirsaran3 & @kujtimprenkuSQA as reviewers. I don't use VS Code enough to understand what plugins are valid other than ESLint. The most important thing to me is Prettier is disabled (in favor of ESLint) as per https://github.com/near-projects/wallet-selector#editor-setup

kujtimprenkuSQA (approved):

rgolea:
Yes, I don't use WebStorm. I'm pretty sure there are configs for each editor we use. Please add the similar configs as best as possible. Maybe add your own editor configurations @lewis-sqa ?

The recommended extensions are just that: recomendations and that can be arbitrary but the configs aren't and disabling prettier from the start is important.

.vscode/extensions.json
```diff
@@ -0,0 +1,18 @@
...
+	// Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+
+	// List of extensions which should be recommended for users of this workspace.
+	"recommendations": [
+		"dbaeumer.vscode-eslint",
+    "kisstkondoros.vscode-codemetrics",
+    "EditorConfig.EditorConfig",
+    "PhilHindle.errorlens",
+    "wix.vscode-import-cost",
+    "syler.sass-indented"
```
lewis-sqa, line 12: Maybe we need ESLint on these files too... Just kidding. Could we format this file correctly though. I think we have a mixture of tabs and spaces!

lewis-sqa (changes requested):

lewis-sqa (approved):

Issue #64, opened 2022-01-21
Signing multiple transactions with Ledger wallet


amirsaran3:
Currently when signing a transaction using the Ledger Wallet option it works the first time. To sign the second time you need to logout and log back in. #23

mikedotexe:
Is this still a problem? Wondering if the code from the MultiSafe helped, like around here:
https://github.com/multi-safe/ui.multisafe/blob/dev/src/near/LedgerSigner.js

(dev branch)

amirsaran3:
I found a weird bug when signing transactions.

I create a transaction with message "1111". Sign it, approve it with ledger. I create another transaction with message "2222". When I try to sign it, on my physical device on the screen I see the arguments from the previous transaction ("1111").

This means that for the second transactions it still using the old buffer data.

I checked the buffer string that I am sending to the Ledger device with APDU logging (using this package `@ledgerhq/logs`) and it has the "2222" message. But still on the screen of the device when approving transaction I see the "1111" message.

This is a possible bug that is causing the error which doesn't let us add messages twice in a row.
```
Uncaught (in promise) Error: Transaction is not signed with the given public key
    at Object.parseRpcError (rpc_errors.js:24:1)
    at json-rpc-provider.js:319:44
    at async Object.exponentialBackoff [as default] (exponential-backoff.js:7:1)
    at async JsonRpcProvider.sendJsonRpc (json-rpc-provider.js:304:26)
```

lewis-sqa:
After some further investigation into this bug, I believe the crux of the problem is we aren't fetching the most recent nonce right before signing. We're currently fetching the once initially and incrementing thereafter. I'm not completely knowledgeable about how nonce tokens work, but in my head the blockchain(?) has moved on by the time we try to sign the next transaction. 

The Multi Safe project referenced above doesn't provide examples of using `sendMessage`. However, the link to `near-ledger-js` was helpful for getting my head around Ledger and how our implementation has come to be 🙂

PR #124, opened 2022-02-10
#64 Sign & Send Transactions (Ledger Wallet)

lewis-sqa:
## This PR Includes

- Refactored `LedgerWallet`
  - Added integration with `LedgerClient`. This abstracts away the details of `@ledgerhq/hw-transport-webhid` and makes it somewhat easier to handle. Unit tests to be written in a follow-up task for the interest of getting this fix over the line.
  - Added `accountId` to `localStorage` to avoid issues with the `getAccount` method.
  - Added support for redirecting to NEAR Wallet if the selected public key is not attached to the account.
  - Added full support for signing transactions with multiple actions.
  - Modified the signing process to fetch the nonce each time instead of only once during initialisation and manually incrementing thereafter.
- Refactored the `<Modal />` component to streamline the Ledger flow, standardise the event handling and clean up the state management.
- Modified the [`finality`](https://docs.near.org/docs/api/rpc#using-finality-param) of the `callFunction` method inside `ProviderService`. This fixes a bug where calling `getMessages` after `sendMessage` would often result in no UI change.
- Fixed the remaining `any` usages and tightened up the ESLint to be an error going forward.
- Created a task to address the lack of unit tests around `LedgerClient`: #126


src/wallets/hardware/LedgerWallet.ts
```diff
@@ -254,95 +200,81 @@ export default class LedgerWallet
...
         keyType: transaction.publicKey.keyType,
         data: signature,
       }),
     });
 
-    const res = await this.provider.sendTransaction(signedTransaction);
-
-    if (typeof res.status !== "string") {
-      const successValue = res.status.SuccessValue || "";
+    return this.provider.sendTransaction(signedTx!).then((res) => {
```
lewis-sqa, line 267: Do we need the `!` here?

lewis-sqa: Need to figure out how to correctly handle `sendTransaction` responses

src/wallets/hardware/LedgerClient.ts
```diff
@@ -0,0 +1,153 @@
...
+}
+
+interface SignParams {
+  data: Uint8Array;
+  derivationPath: string;
+}
+
+interface EventMap {
+  // TODO: Determine event data.
+  disconnect: unknown;
```
lewis-sqa, line 56: Need to test this event to determine the interface

lewis-sqa:
Found a couple of bugs after some more testing:
- State doesn't update if you disconnect from the device instead of the sign out button.
- If you logged in with ledger previously, the page will crash. I think I need to be "lazy" about connecting back to the Ledger. I'm thinking I'll need to prompt at the point of signing to avoid the error.

src/wallets/hardware/LedgerWallet.ts
lewis-sqa: > Need to figure out how to correctly handle `sendTransaction` responses

This really needs addressing across all the wallet implementations.

src/wallets/hardware/LedgerClient.ts
```diff
@@ -0,0 +1,152 @@
...
+  listen(callback: (data: Log) => void) {
+    const unsubscribe = listen(callback);
+
+    return {
+      remove: () => unsubscribe(),
+    };
+  }
+
+  setScrambleKey(key: string) {
+    this.transport.setScrambleKey(key);
```
lewis-sqa, line 78: Well this method does a whole lot of nothing: https://github.com/LedgerHQ/ledgerjs/blob/master/packages/hw-transport-webhid/src/TransportWebHID.ts#L248

amirsaran3 (approved):

PR #125, opened 2022-02-10
Rename package to "near-wallet-selector"

kujtimprenkuSQA:
## This PR includes

- Renamed package from `"near-walletselector"` to `"near-wallet-selector"`.
- Updated `README` to reflect new package name.
- Updated `example `dApp to reflect the new package name.
- Modified `constants ` to be consistent with the new package name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,16 +8,16 @@ The NEAR Wallet Selector makes it easy for users to interact with your dApp. Thi
 
 ## Installation and Usage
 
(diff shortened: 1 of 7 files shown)


lewis-sqa (approved):
@kujtimprenkuSQA, @amirsaran3, @rgolea

It's worth noting that once this is merged, everyone will need to `npm install` in both directories and clear out their local storage.

PR #130, opened 2022-02-14
Added loading indicator when approving ledger public key

amirsaran3:


src/modal/Modal.tsx
```diff
@@ -167,8 +167,12 @@ const Modal: React.FC = () => {
...
                 Dismiss
               </button>
-              <button className="right-button" onClick={handleConnectClick}>
-                Connect
+              <button
+                className="right-button"
+                onClick={handleConnectClick}
+                disabled={state.loading}
+              >
+                {state.loading ? "Loading..." : "Connect"}
```
lewis-sqa, line 175: We should just use `useState` inside the component for whether we're loading. It's purely UI state. We should also disable the dismiss button and make the inputs read-only

lewis-sqa (changes requested):

src/modal/Modal.tsx
amirsaran3: Resolved issues and committed

src/modal/Modal.tsx
```diff
@@ -75,12 +76,16 @@ const Modal: React.FC = () => {
...
+    setIsLoading(true);
     const wallet = state.walletProviders["ledgerwallet"] as ILedgerWallet;
 
     wallet.setDerivationPath(ledgerDerivationPath);
     wallet.setAccountId(ledgerAccountId);
 
-    wallet.signIn().catch((err) => setLedgerError(`Error: ${err.message}`));
+    await wallet
+      .signIn()
+      .catch((err) => setLedgerError(`Error: ${err.message}`));
```
lewis-sqa, line 87: We'll want to set loading to `false` if there's an error too

src/modal/Modal.tsx
```diff
@@ -160,15 +166,24 @@ const Modal: React.FC = () => {
...
                 Dismiss
               </button>
-              <button className="right-button" onClick={handleConnectClick}>
-                Connect
+              <button
+                className="right-button"
+                onClick={handleConnectClick}
+                disabled={isLoading}
+              >
+                {isLoading ? "Loading..." : "Connect"}
```
lewis-sqa, line 186: I know I previously suggested "Loading..." but maybe "Connecting..." would be better here until we have a spinner. Not sure if it causes any reflow problems with it being a longer string?

src/modal/Modal.tsx
lewis-sqa: Ah... We get around this because the `catch` suppresses the Promise rejection

lewis-sqa (approved):

lewis-sqa (changes requested):
Sorry, I did some testing on this branch. We've missed disabling the close button and the backdrop. It causes a situation where I can close the modal and later trigger `signIn` with my Ledger that's waiting for me to approve the public key transaction.

lewis-sqa (approved):

PR #135, opened 2022-02-16
Replacing localStorage with persitent storage

### LINKED ISSUE #134: Implement PersistentStorage instead of localStorage

rgolea:
closes #134


src/services/persistent-storage.service.ts
```diff
@@ -1,11 +1,37 @@
+import { PACKAGE_NAME } from "../constants";
+
 export class PersistentStorage {
   private static instances = new Map<string, PersistentStorage>();
 
   private readonly map = new Map<string, string>();
 
+  static updateInstance(
```
lewis-sqa, line 8: Don't think we need this functionality right now?

src/wallets/helpers.ts
```diff
@@ -1,20 +1,18 @@
...
 import { updateState } from "../state/State";
+import { storage } from "../services/persistent-storage.service";
 
 export const setSelectedWalletId = (walletId: string | null) => {
   if (walletId) {
-    localStorage.setItem(
-      LOCAL_STORAGE_SELECTED_WALLET_ID,
-      JSON.stringify(walletId)
-    );
+    storage.setItem(LOCAL_STORAGE_SELECTED_WALLET_ID, JSON.stringify(walletId));
```
lewis-sqa, line 7: Just tested this and it looks like we're doubling up on the prefix. I noticed the key has changed to `near-wallet-selector-near-wallet-selector:selectedWalletId`. This is happening for all the keys

src/wallets/hardware/LedgerWallet.ts
```diff
@@ -96,9 +97,7 @@ class LedgerWallet implements HardwareWallet {
 
   // TODO: Migrate to storage service (with JSON support).
   private getAuthData = (): AuthData | null => {
-    const authData = localStorage.getItem(
-      LOCAL_STORAGE_LEDGER_WALLET_AUTH_DATA
-    );
+    const authData = storage.getItem(LOCAL_STORAGE_LEDGER_WALLET_AUTH_DATA);
```
lewis-sqa, line 100: It would be really good if we could embed JSON support (hence why it's used for all the values for easy migration) into the storage service. It'll resolve the TODOs and remove class methods like this

lewis-sqa (changes requested):

src/services/persistent-storage.service.ts
```diff
@@ -21,36 +53,42 @@ export class PersistentStorage {
...
       }
     }
   }
 
   clear(): void {
     this.map.clear();
+    this.storage.clear();
   }
+
   getItem(key: string): string | null {
```
lewis-sqa, line 70: We'll need to change the return type for this method now as it won't always be a `string` or `null`. I think we should support something similar to what we have with `.contract.view()` where we can pass a return type:
```ts
// Example Usage:
storage.getItem<AuthData>(LOCAL_STORAGE_LEDGER_WALLET_AUTH_DATA);

// Implementation:
getItem<Value extends unknown>(key: string): Value | null {
  const value = this.map.get(key);

  return typeof value === "undefined"
    ? JSON.parse(this.map.get(key))
    : null;
}

```
This will remove the "@ts-ignore" comment(s) too

lewis-sqa (approved):

PR #144, opened 2022-02-16
Bug fix when reconnecting to ledger after signout

kujtimprenkuSQA:
## This PR includes

- Bug fixed.
- Now when reconnecting after signOut there's no error with the Ledger.

### DIFF
--- a/src/wallets/hardware/LedgerWallet.ts
+++ b/src/wallets/hardware/LedgerWallet.ts
@@ -30,7 +30,7 @@ interface ValidateParams {
 }
 
 class LedgerWallet implements HardwareWallet {
-  private client: LedgerClient;
+  private client: LedgerClient | undefined;
   private subscriptions: Record<string, Subscription> = {};
 
   private provider: ProviderService;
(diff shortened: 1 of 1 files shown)


src/wallets/hardware/LedgerWallet.ts
```diff
@@ -172,9 +172,10 @@ class LedgerWallet implements HardwareWallet {
     if (this.client) {
       await this.client.disconnect();
     }
-
     setSelectedWalletId(null);
     this.emitter.emit("signOut");
+    this.authData = null;
+    this.client = undefined;
```
lewis-sqa, line 178: Might be best if we don't "unset" this variable. The `disconnect` call above should be enough

kujtimprenkuSQA: Even after calling `disconnect` method, `this.client` still will have a value and when we try to reconnect this check is run :

```ts
  private getClient = async () => {
    if (this.client) {
      return this.client;
    }
   //.....
  }
```

If we don't set the value for `this.client` to `undefined `then inside the `validate` method this piece of code throws another error: "The device must be opened first"

```ts
  const publicKey = await client.getPublicKey({
    derivationPath: derivationPath,
  });
```

I hope you understand what I am trying to say since.

lewis-sqa: Ah of course! That `getClient` is a bit of a hack, I must say. Yeah so the issue will be that it doesn't also `connect` so at least this way it will act like a fresh sign in. Nice one 🙂

lewis-sqa: It's technically more correct that it's `LedgerClient | undefined` too 😛

lewis-sqa (approved):

Issue #150, opened 2022-02-17
Convert Example to TypeScript


lewis-sqa:
## Description

Relates to #137.

This task is to convert the `example` project into TypeScript and enforce the same `tsconfig.json` and ESLint configuration as the main project.

PR #152, opened 2022-02-17
#150 Convert Example to TypeScript

lewis-sqa:
## This PR Includes

- Converted `example` project to TypeScript.
- Fixed bug with `on` method not returning `Subscription` object.
- Switched to use ESLint rules from main project.
- Linted `example` project files.


example/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
```
kujtimprenkuSQA, line 3: When I run `"npm run watch"` inside the example it fails with an error `"Reference error React not defined"`

I think we need to add another option ( `"jsx": "react"` ) to the compiler options :


```
  "compilerOptions": {
    "module": "commonjs",
    "jsx": "react"
  },
```


Found this solution here, but there might be other ways to fix it  :
https://github.com/parcel-bundler/parcel/issues/1199

Issue #157, opened 2022-02-22
Fix Sender Wallet Breaking Changes


kujtimprenkuSQA:
## Description

Sender Wallet has been updated and they have introduced some breaking changes that make our integration fail.
We are not yet sure what are the current changes since there's no public documentation regarding this update.
We might need to ask on discrord if we get blocked on any functionality:

- Follow this example: https://github.com/SenderWallet/sender-wallet-dapp-example

## Acceptance Criteria

- Make sure that all the current functionality works as it used to before this update.

kujtimprenkuSQA:
The Sender Wallet team is still working on the extension. 
This task needs to be moved in **"TODO"** column until they release a stable version and have some proper documentation.

PR #162, opened 2022-02-25
feat: sender wallet API change

think-in-universe:
Fixes: #157 

Due to the recent Sender API change (as documented here: https://docs.senderwallet.io/api-reference/near-provider-api), `near-wallet-selector` needs to be updated to work well with the latest version. 

A relevant idea is to maintain versioned implementations which could be loaded, by detecting the version of the installed extension wallet, but this is not high-priority since we prefer to standardize and stabilize the wallet SDKs.

### DIFF
--- a/src/interfaces/InjectedSenderWallet.ts
+++ b/src/interfaces/InjectedSenderWallet.ts
@@ -41,8 +41,6 @@ export interface RequestSignInParams {
   amount?: string; // in yoctoⓃ
 }
 
-export type Callback = (response: any) => void;
-
 export interface RpcChangedResponse {
   method: "rpcChanged";
   notificationId: number;
@@ -86,6 +84,13 @@ export interface RequestSignTransactionsParams {
   transactions: Array<Transaction>;
 }
 
+export interface SenderWalletEvents {
+  signIn: () => void;
+  signOut: () => void;
+  accountChanged: (changedAccountId: string) => void;
+  rpcChanged: (response: RpcChangedResponse) => void;
+}
+
 interface InjectedSenderWallet {
   isSender: boolean;
   getAccountId: () => string;
@@ -95,7 +100,10 @@ interface InjectedSenderWallet {
   ) => Promise<RequestSignInResponse>;
   signOut: () => boolean;
   isSignedIn: () => boolean;
-  on: (event: string, callback: Callback) => void;
+  on: <Event extends keyof SenderWalletEvents>(
+    event: Event,
+    callback: SenderWalletEvents[Event]
+  ) => void;
   // TODO: Determine return type.
   sendMoney: (params: SendMoneyParams) => Promise<unknown>;
   signAndSendTransaction: (
--- a/src/interfaces/InjectedSenderWallet.ts
+++ b/src/interfaces/InjectedSenderWallet.ts
@@ -40,11 +40,6 @@ export interface RequestSignInParams {
   methodNames?: Array<string>;
 }
 
-export interface SignOutResponse {
-  // TODO: Figure out when this isn't "success".
-  result: "success";
-}
-
 export type Callback = (response: any) => void;
 
 export interface RpcChangedResponse {
@@ -91,13 +86,13 @@ export interface RequestSignTransactionsParams {
 }
 
 interface InjectedSenderWallet {
-  isSender: string,
+  isSender: boolean,
   getAccountId: () => string;
   getRpc: () => Promise<GetRpcResponse>;
   requestSignIn: (
     params: RequestSignInParams
   ) => Promise<RequestSignInResponse>;
-  signOut: () => SignOutResponse;
+  signOut: () => boolean;
   isSignedIn: () => boolean;
   on: (event: string, callback: Callback) => void;
   // TODO: Determine return type.
--- a/src/interfaces/InjectedSenderWallet.ts
+++ b/src/interfaces/InjectedSenderWallet.ts
@@ -38,6 +38,7 @@ export interface GetRpcResponse {
 export interface RequestSignInParams {
   contractId: string;
   methodNames?: Array<string>;
+  amount?: string; // in yoctoⓃ
 }
 
 export type Callback = (response: any) => void;
--- a/src/interfaces/InjectedSenderWallet.ts
+++ b/src/interfaces/InjectedSenderWallet.ts
@@ -1,9 +1,5 @@
 // Interfaces based on "documentation": https://github.com/SenderWallet/sender-wallet-integration-tutorial
 
-export interface InitParams {
-  contractId: string;
-}
-
 // Empty string if we haven't signed in before.
 interface AccessKey {
   publicKey: {
@@ -11,10 +7,6 @@ interface AccessKey {
     keyType: number;
   };
   secretKey: string;
-}
-
-export interface InitResponse {
-  accessKey: AccessKey | "";
 }
 
 export interface RequestSignInResponse {
@@ -53,7 +45,7 @@ export interface SignOutResponse {
   result: "success";
 }
 
-export type AccountChangedCallback = (newAccountId: string) => void;
+export type Callback = (response: any) => void;
 
 export interface RpcChangedResponse {
   method: "rpcChanged";
@@ -61,8 +53,6 @@ export interface RpcChangedResponse {
   rpc: RpcInfo;
   type: "sender-wallet-fromContent";
 }
-
-export type RpcChangedCallback = (newRpc: RpcChangedResponse) => void;
 
 export interface SendMoneyParams {
   receiverId: string;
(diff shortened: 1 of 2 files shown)


src/wallets/injected/SenderWallet.ts
```diff
@@ -92,14 +88,9 @@ class SenderWallet implements InjectedWallet {
...
 
-    const rpcResponse = await this.wallet.getRpc();
-
-    if (!this.networkMatches(rpcResponse)) {
-      return;
-    }
-
     const { accessKey } = await this.wallet.requestSignIn({
       contractId: this.options.contract.accountId,
+      methodNames: this.options.contract.changeMethods || [],
```
think-in-universe, line 93: Is `methodNames` needed? 

It impacts whether the same function call key can be reused or not, once some new methodsNames are added or removed.

lewis-sqa: I was experimenting with the options API for contracts but hadn't implemented `changeMethods` or `viewMethods`.

I did some digging and found both Sender Wallet and NEAR Wallet use `methodNames` instead of having two parameters. I'd omit this change for now and we can raise another PR in the future that makes use of this option for all the wallets to avoid inconsistency.

src/interfaces/InjectedSenderWallet.ts
```diff
@@ -101,16 +91,15 @@ export interface RequestSignTransactionsParams {
 }
 
 interface InjectedSenderWallet {
-  init: (params: InitParams) => Promise<InitResponse>;
+  isSender: string,
```
lewis-sqa, line 94: I believe this property returns `boolean`

src/interfaces/InjectedSenderWallet.ts
```diff
@@ -101,16 +91,15 @@ export interface RequestSignTransactionsParams {
...
   getAccountId: () => string;
   getRpc: () => Promise<GetRpcResponse>;
   requestSignIn: (
     params: RequestSignInParams
   ) => Promise<RequestSignInResponse>;
-  signOut: () => Promise<SignOutResponse>;
+  signOut: () => SignOutResponse;
```
lewis-sqa, line 100: This method seems to return a `boolean` now

src/interfaces/InjectedSenderWallet.ts
```diff
@@ -101,16 +91,15 @@ export interface RequestSignTransactionsParams {
...
   isSignedIn: () => boolean;
-  onAccountChanged: (callback: AccountChangedCallback) => void;
-  onRpcChanged: (callback: RpcChangedCallback) => void;
+  on: (event: string, callback: Callback) => void;
```
lewis-sqa, line 102: To address the `any` issue (this project forbids them), we should look to create an event map to ensure callback methods have the correct types:

```ts
export interface SenderWalletEvents {
  signIn: () => void;
  signOut: () => void;
  accountChanged: (changedAccountId: string) => void;
  rpcChanged: (response: RpcChangedResponse) => void;
}

on: <Event extends keyof SenderWalletEvents>(
    event: Event,
    callback: SenderWalletEvents[Event]
  ) => void;
```

src/interfaces/InjectedSenderWallet.ts
```diff
@@ -101,16 +91,15 @@ export interface RequestSignTransactionsParams {
 }
 
 interface InjectedSenderWallet {
-  init: (params: InitParams) => Promise<InitResponse>;
+  isSender: string,
   getAccountId: () => string;
   getRpc: () => Promise<GetRpcResponse>;
   requestSignIn: (
```
lewis-sqa, line 97: I'm rather concerned about the current state of this method since the update. I'm not sure if I'm doing something wrong but simply running the following always resolves:
```ts
near.requestSignIn({ contractId: "guest-book.testnet", methodNames: [] })
  .then((x) => console.log("then", x))
  .catch((err) => console.log("catch", err.message));
```
It logs the following to console: `"then Please connect first"`. There's three issues with this really:
- What exactly do I need to connect first? My wallet is unlocked etc.
- This is really an error so I should really see the following in console: `"catch Please connect first"`.
- The return type is either a `string` or an `object` now?

lewis-sqa: It looks like we can also pass `amount` as a parameter to this method?

src/wallets/injected/SenderWallet.ts
```diff
@@ -56,7 +56,7 @@ class SenderWallet implements InjectedWallet {
   };
 
   private isInstalled = () => {
-    return !!window.wallet;
+    return typeof window.near !== 'undefined' && window.near.isSender;
```
lewis-sqa, line 59: We could be more terse on this expression with:
```ts
return !!window.near?.isSender;
```

lewis-sqa (commented):
Hey @think-in-universe,

Thanks for your efforts on this PR! I've left a few comments after pulling down your fork and enabling our workflow which has highlighted a few linting issues

src/interfaces/InjectedSenderWallet.ts
```diff
@@ -101,16 +91,15 @@ export interface RequestSignTransactionsParams {
...
   ) => Promise<RequestSignInResponse>;
-  signOut: () => Promise<SignOutResponse>;
+  signOut: () => SignOutResponse;
   isSignedIn: () => boolean;
-  onAccountChanged: (callback: AccountChangedCallback) => void;
-  onRpcChanged: (callback: RpcChangedCallback) => void;
+  on: (event: string, callback: Callback) => void;
   // TODO: Determine return type.
   sendMoney: (params: SendMoneyParams) => Promise<unknown>;
   signAndSendTransaction: (
```
lewis-sqa, line 105: It looks like the multi-transaction method, `requestSignTransactions` has been renamed to `signAndSendTransactions` when testing the extension directly in my browser. This contradicts the docs: https://docs.senderwallet.io/api-reference/near-provider-api#near.requestsigntransactions

src/interfaces/InjectedSenderWallet.ts
think-in-universe: `Please connect first`: This is an unexpected behavior from the wallet which should be fixed in next release (say next Monday). There's a `connect()` API added previously which needs to be called before `requestSignIn()`, but I have suggested to remove that.

think-in-universe:
Please notice that this PR here doesn't work with the current version of Sender wallet, which contains some break change to the API doc https://docs.senderwallet.io/api-reference/near-provider-api. But the PR is expected to work with the upcoming release that should would be ready early next week, which should follow the API standard.

src/interfaces/InjectedSenderWallet.ts
think-in-universe: That's true. The next release should fix the the API inconsistency issue, which should be ready earlier but it's pending on review by Chrome Web Store.

src/wallets/injected/SenderWallet.ts
think-in-universe: sounds good

src/wallets/injected/SenderWallet.ts
think-in-universe: yes. `methodNames` are used in `requestSignIn()`, and `changeMethods` / `viewMethods` are used when initializing the contract instance.

src/interfaces/InjectedSenderWallet.ts
think-in-universe: yes. thanks.

src/interfaces/InjectedSenderWallet.ts
think-in-universe: great. will add the event map. .

src/interfaces/InjectedSenderWallet.ts
lewis-sqa: > `Please connect first`: This is an unexpected behavior from the wallet which should be fixed in next release (say next Monday or Tuesday). There's a `connect()` API added previously which needs to be called before `requestSignIn()`, but I have suggested to remove that.

Ah that makes sense. I've now noticed the `connect` method commented out in the example project and when checking directly in the browser. The main things for me are the last 2 points. I'm worried we'll either get runtime errors because the types are mismatched or the Promise is resolving when in fact there's an error, so the wallet selector won't detect something has gone wrong.

> The `amount` here means the `allowance` of the function call key, which is 0.25N by default.

Sorry, I was mostly referring to the types for this. We should add the `amount` property to `RequestSignInParams` for completeness. It's possible we might want to add support for this when we address the `methodNames` option too

think-in-universe: I have added the parameter type of `amount`. It's a string in yoctoⓃ

think-in-universe:
@lewis-sqa tested this PR with the latest v0.1.4 version Sender and it's working well now.

src/wallets/injected/SenderWallet.ts
```diff
@@ -92,12 +88,6 @@ class SenderWallet implements InjectedWallet {
       await this.init();
     }
 
-    const rpcResponse = await this.wallet.getRpc();
-
-    if (!this.networkMatches(rpcResponse)) {
-      return;
-    }
```
lewis-sqa, line 99: How do we now ensure the `networkId` matches the dApp?

lewis-sqa (approved):

PR #169, opened 2022-02-28
Add Angular Example

kujtimprenkuSQA:
## This PR includes
- Guest book example with the Angular framework.
- Same UI as the react example.
- Using the already deployed contract `guest-book.testnet`
- Up to date with the latest changes of `dev` branch.
- Successfuly add message with donation.

## Work remaining
- [ ] Investigate error when sending messages without donation.
- [ ] Remove /example
- [ ] Update favicon and index.html tags.


examples/angular/src/app/components/content/content.component.html
```diff
@@ -0,0 +1,15 @@
+<ng-container>
+  <div *ngIf="account">
+    <button (click)=signOut()>Log out</button>
+    <button (click)=switchProvider()>Switch Provider</button>
+  </div>
+  <app-form *ngIf="account" [account]="account" (onAddMessage)="onSubmit($event)"></app-form>
+  <app-messages *ngIf="!!account && !!messages.length" [messages]="messages"></app-messages>
+</ng-container>
```
lewis-sqa, line 8: I'm a complete novice with Angular but are we able to just do a single `ngIf` on the container for `account`? Then `app-messages` can just check `messages.length`?

examples/angular/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "angular",
```
lewis-sqa, line 2: To align with the other examples, could we name this `near-wallet-selector-example-angular`

examples/angular/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "@angular/animations": "~13.2.0",
+    "@angular/common": "~13.2.0",
+    "@angular/compiler": "~13.2.0",
+    "@angular/core": "~13.2.0",
+    "@angular/forms": "~13.2.0",
+    "@angular/platform-browser": "~13.2.0",
+    "@angular/platform-browser-dynamic": "~13.2.0",
+    "@angular/router": "~13.2.0",
+    "near-wallet-selector": "*",
+    "process": "^0.11.10",
```
lewis-sqa, line 22: Little concerned about using this module as it hasn't changed in 5 years! Are we able to simply polyfill process to an empty object or maybe `{ env: {} }`?

examples/angular/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "@angular/cli": "~13.2.4",
+    "@angular/compiler-cli": "~13.2.0",
+    "@types/jasmine": "~3.10.0",
+    "@types/node": "^12.11.1",
+    "jasmine-core": "~4.0.0",
+    "karma": "~6.3.0",
+    "karma-chrome-launcher": "~3.1.0",
+    "karma-coverage": "~2.1.0",
+    "karma-jasmine": "~4.0.0",
+    "karma-jasmine-html-reporter": "~1.7.0",
```
lewis-sqa, line 38: I removed all the test frameworks in the other examples to keep these projects super minimum. It might be worth doing the same for this?

lewis-sqa (commented):
Just a few minor point from me! 👍 I love how similar yet different the 3 frameworks are! 😛

examples/angular/package.json
kujtimprenkuSQA: I tried to do it this way but for some reason I could not get it working and when I tried to send messages without a "donation" `near-api-js` was throwing a strange error which in the end I found it was looking for `process.env.SOME_VARIABLE` and I lost a few hours thinking env variables were working properly :(

examples/angular/package.json
kujtimprenkuSQA: I already removed test files, I will remove the packages too.

examples/angular/package.json
kujtimprenkuSQA: Thanks, did not notice it :)

examples/angular/src/app/components/content/content.component.html
kujtimprenkuSQA: Sure it can work that way too, I just left the markup this way to keep it somehow separated!

Issue #164, opened 2022-02-27
Enhancement: API that sends multiple transactions 


think-in-universe:
The current version, `near-wallet-selector` supports [`signAndSendTransaction`](https://github.com/near-projects/wallet-selector/blob/06b22086996dc795372969240ee25910fb61b014/src/wallets/Wallet.ts#L58-L60) which sends a single transaction.

I received feedback that developers want to send multiple transactions in some cases. Is this possible to be supported in `near-wallet-selector`?

lewis-sqa:
Ran into a few issues detailed in #170. I've moved this task into blocked until it's clear exactly how it will work for the various wallet implementations.

mikedotexe:
hey @think-in-universe can you help me understand the use case for someone doing multiple transactions?

PR #170, opened 2022-03-01
#164 Add Support for Batch Transactions

lewis-sqa:
## This PR Includes

- Added support for `signAndSendTransactions` NEAR Wallet & Sender Wallet.

## Notes

- I'm struggling to understand how we might implement this for Ledger Wallet.
  - The `near-api-js` docs seem to [confuse Transactions with Actions](https://docs.near.org/docs/api/naj-cookbook#batch-transactions).
  - There's [no dedicated RPC method for multiple transactions](https://near.github.io/near-api-js/classes/providers_json_rpc_provider.jsonrpcprovider.html#sendtransaction). Surely it would defeat the purpose of batch transactions if we simply call `sendTransaction` sequentially - we would lack atomicity.
- [Sender Wallet only supports `receiverId` and `actions` for Transactions](https://docs.senderwallet.io/guide/sending-transactions#sign-and-send-multiple-transactions) compared to NEAR Wallet. There's again (like `signAndSendTransaction`) the limitation of only `FunctionCall` Actions too.
- NEAR Wallet [doesn't define `requestSignTransactions` in TypeScript](https://github.com/near/near-api-js/blob/7ea21f330af1a00543b1ae655761a98820f6a368/src/wallet-account.ts#L280) although it's documented [here](https://near.github.io/near-api-js/classes/walletaccount.walletconnection.html#requestsigntransactions).
- The tests in `near-api-js` [only demonstrate a single transaction](https://github.com/near/near-api-js/blob/c73eb80c5a6b5fb4f5e8105fc02db2d8e8c01a1d/test/wallet-account.test.js#L163) which makes me think this feature is a bit of an edge case, especially when the docs only demonstrate multiple Actions too.

## Work Remaining

- [ ] Test NEAR Wallet.
- [ ] Test Sender Wallet.
- [ ] Update README with new method.

### DIFF
--- a/src/core/Contract.ts
+++ b/src/core/Contract.ts
@@ -3,7 +3,10 @@ import ProviderService, {
 } from "../services/provider/ProviderService";
 import WalletController from "../controllers/WalletController";
 import { Options } from "./NearWalletSelector";
-import { SignAndSendTransactionParams } from "../wallets/Wallet";
+import {
+  SignAndSendTransactionParams,
+  SignAndSendTransactionsParams,
+} from "../wallets/Wallet";
 
 class Contract {
   private readonly options: Options;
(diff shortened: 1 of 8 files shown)


lewis-sqa:
Completed as part of #231, #239, #240 & #241.

Issue #173, opened 2022-03-02
Unify Sign In Functionality Across Wallets


lewis-sqa:
## Description

The current approach to sign in is less than ideal when Ledger (or more broadly speaking, hardware) wallet is in the picture. There are 3 areas I want to address in this task:

- Impure `signIn` method for Ledger wallet [here](https://github.com/near-projects/wallet-selector/blob/a3f9abeb80a4ebc40ccdce57d490364d17561f8b/src/wallets/hardware/LedgerWallet.ts#L109-L128).
- Awkward logic in `<Modal />` [here](https://github.com/near-projects/wallet-selector/blob/a3f9abeb80a4ebc40ccdce57d490364d17561f8b/src/modal/Modal.tsx#L102-L111).
- Unable to call [`signIn` programmatically](https://github.com/near-projects/wallet-selector/blob/a3f9abeb80a4ebc40ccdce57d490364d17561f8b/src/core/NearWalletSelector.tsx#L87-L89) via top-level API for Ledger Wallet.

A new approach is to convert the parameters of `signIn` into a dynamic object for the given wallet's type. This should remove the need for `setDerivationPath` and `setAccountId` on HardwareWallet:

```ts
// NEAR Wallet (BrowserWallet)
await selector.signIn({ walletId: "near-wallet" });

// Sender Wallet (InjectedWallet)
await selector.signIn({ walletId: "sender-wallet" });

// Ledger Wallet (HardwareWallet)
await selector.signIn({
  walletId: "ledger-wallet",
  accountId: "account-id.testnet",
  derviationPath: "44'/397'/0'/0'/1'",
});
```

PR #178, opened 2022-03-03
#173 Unify Sign In Functionality Across Wallets

lewis-sqa:
## This PR Includes

- `signIn` now takes hardware wallet specific parameters, removing the need for additional state in `LedgerWallet` and `setAccountId` and `setDerivationPath`.
- Updated README to reflect changes.
- Exported library to the `window` in each example for "headless" testing such as this `signIn` method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Sign in (programmatically):
 ```ts
 // NEAR Wallet.
 await selector.signIn({
- walletId: "near-wallet"
+  walletId: "near-wallet"
 });
 
 // Sender Wallet.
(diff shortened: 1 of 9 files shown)


kujtimprenkuSQA (commented):
Can we rename const  `near` to `selector` as it's done in the above documentation just to be consistent:
From: 
```ts
const near = new NearWalletSelector({
  ...otherOptions,
  theme: "light",
});
```
To:

```ts
const selector = new NearWalletSelector({
  ...otherOptions,
  theme: "light",
});
```

The lines which need change are line: 161, 169 and 183.

Perhaps you have addressed this in another PR which I still haven't had a look at if so just ignore this comment :)

lewis-sqa:
> Perhaps you have addressed this in another PR which I still haven't had a look at if so just ignore this comment :)

@kujtimprenkuSQA Yeah that's correct 🙂 #179 removes the need for this section as it's covered in a new "Options" section

kujtimprenkuSQA (approved):

PR #181, opened 2022-03-04
[POC] Wallet Modules

lewis-sqa:
## This PR Includes

- Switch from built-in wallet IDs to direct imports. This allows future improvements for wallets that require specific settings without breaking the polymorphic architecture.
- Added support for custom wallets without us becoming blockers with approving PRs etc. The required interfaces are exported in the entry file.


rgolea:
I would rather separate them in packages. It's easier to make sure the code is not in the code and there is no tree-shaking required to be implemented in the project the wallet selector ends up in.

lewis-sqa:
To get most of the value in this PR, I've pulled back the breaking changes initially implemented (direct imports instead of `walletId` strings). There's still some debate on how we should expose the wallet implementations (separate packages, namespacing with code-splitting support etc).

PR #182, opened 2022-03-04
Fix UI text and border colors

kujtimprenkuSQA:
## This PR includes

- Fix: text color is now applied properly when no theme is provided in options.

## Notes

Since it was just 1 commit I did not open an issue for it.
The problem was when no theme is provided and the OS/Browser is configured to use system's Light mode then the text color was not applied properly, and also the border of buttons was not applied properly.

### DIFF
--- a/src/modal/Modal.styles.ts
+++ b/src/modal/Modal.styles.ts
@@ -32,7 +32,7 @@ export default `
   box-sizing: content-box;
 }
 
-.Modal.Modal-light-theme {
+.Modal, .Modal.Modal-light-theme {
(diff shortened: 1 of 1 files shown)


lewis-sqa (approved):
Ah my bad. I originally didn't think we defaulted to the system settings but took the README as gospel... I should have tested it, sorry! 🙁

kujtimprenkuSQA:
You actually did nothing wrong here, this was probably a bug form way back when I worked in styles it just happened that I noticed it now :)

PR #184, opened 2022-03-07
Added LedgerClient.spec.ts

amirsaran3:


src/wallets/hardware/LedgerClient.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+
+afterEach(() => {
+  jest.resetModules();
+});
+
+describe("getVersion", () => {
+  it("returns the current version", async () => {
+    const { client, transport } = createLedgerClient({
+      transport: {
+        // TODO: Examine real response so this mock is reliable.
```
lewis-sqa, line 38: We'll want to have a look at these responses in the real implementation and mimic them. I tend to extract these out as "fixture helpers". The detail of them isn't terribly important usually so I do something like this to keep the tests clean (again don't forget about the real responses - my template on Friday was just a rough approach):

```ts
// ledgerWalletMocks.ts
const createGetVersionResponseMock = () => {
  // TODO: Examine real response so this mock is reliable.
  return Buffer.from([1, 2, 3]);
};

// LedgerClient.spec.ts
const { client, transport } = createLedgerClient({
  transport: {
    send: jest.fn().mockResolvedValue(
      createGetVersionResponseMock()
    ),
  },
});

File: src/wallets/hardware/LedgerClient.spec.ts [hunk 2]
```diff
@@ -0,0 +1,167 @@
...
+        send: jest.fn().mockResolvedValue(Buffer.from([1, 2, 3])),
+      },
+    });
+
+    await client.connect();
+    const result = await client.getVersion();
+
+    expect(transport.send).toHaveBeenCalledWith(128, 6, 0, 0);
```
lewis-sqa, line 46: We could export the constants from LedgerClient.ts and reference them here so the tests have a bit more meaning than seemingly random numbers

src/wallets/hardware/LedgerClient.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+    });
+
+    expect(transport.send).toHaveBeenCalledWith(
+      128,
+      4,
+      0,
+      87,
+      // TODO: We should really expect a correctly formatted buffer.
+      //  basically we'll be testing that 'parseDerivationPath' works as expected.
+      expect.any(Buffer)
```
lewis-sqa, line 72: I was a bit lazy here using `expect.any`. As the TODO suggests, we'll want to be more precise so we can test the `parseDerivationPath` method is working correctly

src/wallets/hardware/LedgerClient.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+        115, 116, 110, 101, 116, 0, 226, 125, 56, 106, 199, 195, 73, 246, 10,
+        249, 57, 121, 249, 233, 201, 22, 102, 15, 131, 165, 129, 76, 109, 40,
+        170, 241, 102, 140, 43, 133, 200, 31, 2, 146, 147, 11, 47, 69, 0, 0, 18,
+        0, 0, 0, 103, 117, 101, 115, 116, 45, 98, 111, 111, 107, 46, 116, 101,
+        115, 116, 110, 101, 116, 211, 174, 138, 47, 84, 199, 21, 141, 240, 194,
+        235, 66, 156, 178, 230, 10, 240, 60, 202, 182, 118, 176, 120, 45, 97,
+        184, 167, 79, 177, 64, 149, 228, 1, 0, 0, 0, 2, 10, 0, 0, 0, 97, 100,
+        100, 77, 101, 115, 115, 97, 103, 101, 14, 0, 0, 0, 123, 34, 116, 101,
+        120, 116, 34, 58, 34, 97, 115, 100, 34, 125, 0, 224, 87, 235, 72, 27, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
lewis-sqa, line 100: We might be able to clean this input up by actually using `near-api-js` 🤔 and creating a simple Transaction? The assertions probably need some comments and use of the constants I mentioned previously. For example the first send call is to clear the buffer by retrieving the version; the rest are the chunks related to the transaction

src/wallets/hardware/LedgerClient.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+    const { client, transport } = createLedgerClient({
+      transport: {
+        off: jest.fn(),
+      },
+    });
+
+    await client.connect();
+    await client.off("connect", jest.fn());
+    expect(transport.off).toHaveBeenCalledWith("connect", expect.any(Function));
+    expect(transport.off).toHaveBeenCalledTimes(1);
```
lewis-sqa, line 150: Somewhat similar to my comments on `.on`, we can omit the off mock here as `jest-mock-extended` has us covered for this. We mock `send` because it requires custom logic which it can't infer from types alone - it's basically a dynamic mock. Here we just want to assert the mocked function is called a particular way and not the value it returns etc

lewis-sqa (approved):

PR #156, opened 2022-02-22
Automating support for npm publishing

rgolea:
This is a first draft. I need some feedback.

Closes #109 #138

PR #196, opened 2022-03-11
Attempt to convert to a monorepo.

### LINKED ISSUE #137: Convert to a Monorepo
## Description

Our current setup of the project makes it difficult to manage both the library and the example integration. By moving to a monorepo, we will be able to take advantage of [TypeScript Paths](https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping) and remove our hack with `npm run watch`.

Other perks to this setup means we could eventually extract the different wallet implementations to create a plugin system not too dissimilar to Webpack, Passport, ESLint and Rollup.

The two suggested approaches are: [Lerna](https://lerna.js.org/) and [Nx](https://nx.dev/).

## Acceptance Criteria

- We can build and publish `near-wallet-selector` separately to the other packages (e.g. example integration which will be [private](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#private)).
- We are able to run the example integration (using TypeScript Paths) as though it's importing from NPM but in fact using the local package (and removes the need symlinks and nodemon + 2 build commands 😩).
- We can use Yarn again (`prepare` hook only worked with NPM). 
- NPM scripts are updated to reflect new project setup.

rgolea:
This should be tested a lot first and we should also merge #156 before we close go ahead.

closes #137

Vue will not work for now like this. See [#222](https://github.com/ZachJW34/nx-plus/issues/222)


.vscode/extensions.json
```diff
@@ -1,14 +1,9 @@
 {
-  // See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
-  // Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
-  // List of extensions which should be recommended for users of this workspace.
   "recommendations": [
     "dbaeumer.vscode-eslint",
     "EditorConfig.EditorConfig",
-    "syler.sass-indented"
+    "syler.sass-indented",
+    "firsttris.vscode-jest-runner"
```
lewis-sqa, line 6: Maybe for another PR but it might be useful to suggest the NX plugin in here

packages/near-wallet-selector/src/wallets/injected/SenderWallet.ts
```diff
@@ -82,7 +83,7 @@ function setupSenderWallet(): WalletModule<InjectedWallet> {
             await this.signOut();
             await this.signIn();
           } catch (e) {
-            logger.log(`Failed to change account ${e.message}`);
+            logger.log(`Failed to change account ${(e as unknown as Error).message}`);
```
lewis-sqa, line 86: Was this an auto fix? 😆 I'll let some of this slide as we're aware the ESLint configuration needs to be reintegrated maintaining compatibility with NX

lewis-sqa (commented):
Really great work here! 🎉 As discussed yesterday, there are future improvements to be made but this PR is super valuable in it's current state.

I think the only outstanding issue is the pipeline side. I think the build command is broke at the moment. Do we need to switch them to `nx` commands? Not sure if the same applies for tests? Is there a concept of "build all the things" and "test all the things"? 😛

lewis-sqa:
@rgolea could we fix the PR Actions workflow so this can be merged?

rgolea:
@lewis-sqa #203 should have fixed that. If you need to see those changes separately, please see there.

package.json
```diff
@@ -2,74 +2,118 @@
...
+  },
+  "license": "ISC",
+  "author": "near",
   "main": "./lib/cjs/index.js",
   "module": "./lib/esm/index.js",
   "types": "./lib/esm/index.d.ts",
   "scripts": {
-    "build:esm": "rm -rf lib/esm && tsc --module esnext --outDir lib/esm",
-    "build:cjs": "rm -rf lib/cjs && tsc --module commonjs --outDir lib/cjs",
     "build": "yarn build:esm && yarn build:cjs",
```
lewis-sqa, line 23: This needs updating to use nx

package.json
```diff
@@ -2,74 +2,118 @@
...
+    "build:cjs": "rm -rf lib/cjs && tsc --module commonjs --outDir lib/cjs",
+    "build:esm": "rm -rf lib/esm && tsc --module esnext --outDir lib/esm",
     "lint": "eslint ./src --ext .ts,.tsx",
     "lint:fix": "yarn lint --fix",
+    "prepack": "yarn build",
     "test": "jest",
```
lewis-sqa, line 29: This also needs updating to use nx

lewis-sqa (approved):

PR #202, opened 2022-03-17
React example fix

mikedotexe:
# Description

As discussed in Discord, I was seeing this error when trying to run the React example on the `main` branch.

image

I also used changed the dependency for the `near-wallet-selector` to use a relative path.

## Broken behavior

I created a new directory, let's call it `~/tmp` and I cloned the `near-wallet-selector` repo. By default this checked out the `dev` branch. In the project root I ran `yarn && yarn build` then `cd examples/react` and `yarn && yarn start`

Then I switched branches to `main` went back to the project root and again ran `yarn && yarn build` then `cd examples/react` and `yarn && yarn start` and saw the message in the screenshot.

## After this PR

After this PR you can follow the same procedure and not see an error message and have it work as expected.


lewis-sqa:
@mikedotexe I'm a little confused about this PR. I'm guessing you've based your changes on `main` as it's 81 commits behind `dev` where we already have this change to the `package.json` (as well as in the `angular` and `vue` examples). Maybe it was an issue with `parcel` not picking up the changes when you switched branches and rebuilt (but didn't restart the bundler)?

In terms of the change to `Modal.tsx`, maybe I'm having a slow day but I believe the code more or less works the same way but just written differently? `options.ui?.description` would give you `undefined` without crashing if `ui` is not there. It then "defaults" using the OR operator to display our fallback string if it's `undefined` or an empty string.

Screenshot 2022-03-17 at 09 28 33

I'm certainly not suggesting our build system is perfect - it's actually a bit of a pain imo though I'm hopeful the new monorepo changes will help with this. While it isn't ideal, I'd suggest a clean build whenever doing something as drastic as changing branches right now as we're moving pretty quick with changes. A future improvement might be that we add a `clean` or `setup` command that essentially deletes the all caches (including `node_modules`) and rebuilds everything to avoid this kind of issue

PR #247, opened 2022-04-08
fix: Fixed bug where error was not shown when connecting with ledger

amirsaran3:
# Description
If you want to connect using the Ledger Wallet and you type in an account id that is not registered with the public key an error will not be shown. There will not be any UI changes or messages to inform you that something went wrong.

This is happening because after we update the `ledgerError` state variable we also reset the state using the `resetState` method.

Now the `resetState` method will only be called when `state.showModal` changes using the `useEffect` hook.

# Checklist:

# Breaking changes

### DIFF
--- a/packages/core/src/lib/modal/Modal.tsx
+++ b/packages/core/src/lib/modal/Modal.tsx
@@ -50,6 +50,10 @@ const Modal: React.FC<ModalProps> = ({ options, network, wallets }) => {
     };
   }, []);
 
+  useEffect(() => {
+    resetState();
+  }, [state.showModal]);
+
   const resetState = () => {
     setWalletInfoVisible(false);
     setLedgerError("");
@@ -67,8 +71,6 @@ const Modal: React.FC<ModalProps> = ({ options, network, wallets }) => {
       ...prevState,
       showModal: false,
     }));
-
-    resetState();
   };
 
   const handleDismissOutsideClick = (e: MouseEvent) => {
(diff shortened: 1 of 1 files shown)


packages/core/src/lib/modal/Modal.tsx
```diff
@@ -50,6 +50,10 @@ const Modal: React.FC<ModalProps> = ({ options, network, wallets }) => {
     };
   }, []);
 
+  useEffect(() => {
```
kujtimprenkuSQA, line 53: _Not sure if the PR is ready for review but I have a some suggestions:_

I think we could just call `resetState()` at the end of the `handleDismissClick()` function.
This would avoid unnecessarily reseting the state in 2 cases, with `useEfect` the state is being reset when the component first loads and when we first open the modal.

And maybe it would be nice to clear the ledgerError  `setLedgerError("");` as soon as we click the "Connect" `handleConnectClick()` .
For example let's say I provided an invalid account id after I click connect and confirm it in ledger then the error will be shown at this moment let say I made  a typo in the account id after I fix it and click the connect button again it would be nice to clear the error until ledger throws the next error if there's any.

amirsaran3: The problem with having the `resetState()` function just inside the `handleDismissClick()` is that when we don't click dismiss and click connect and everything goes correctly. In this case we did not reset the state. Now if we disconnect and try to connect using ledger again the inputs will have the old state data.

The reason we can't just add `resetState()` inside the `handleConnectClick()` is that the user will not see the errors after we reset the state.

kujtimprenkuSQA: Yeah that seems to be the case.

kujtimprenkuSQA (approved):

PR #256, opened 2022-04-13
feat: Removed account id requirement for ledger connecting

amirsaran3:
# Description

Removed the need for users to input their account id when connecting with the Ledger Wallet. It now automatically fetches the account id connected with your public key you are connecting with and validates it. If there is no account id connected with public key you are trying to connect with it will show you an error message.

# Checklist:

# Breaking changes

### DIFF
--- a/packages/ledger/src/lib/ledger.ts
+++ b/packages/ledger/src/lib/ledger.ts
@@ -18,7 +18,11 @@ interface AuthData {
 
 interface ValidateParams {
   accountId: string;
-  derivationPath: string;
+  publicKey: string;
+}
+
+interface GetAccountIdFromPublicKeyParams {
+  publicKey: string;
 }
 
 interface LedgerState {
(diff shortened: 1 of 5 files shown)


packages/ledger/src/lib/ledger.ts
```diff
@@ -141,6 +139,41 @@ export function setupLedger({
...
+      publicKey,
+    }: GetAccountIdFromPublicKeyParams): Promise<string> => {
+      const response = await fetch("https://rest.nearapi.org/explorer", {
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify({
+          user: "public_readonly",
+          host: "testnet.db.explorer.indexer.near.dev",
```
kujtimprenkuSQA, line 152: I am actually not familiar with this but will this work on a **`.near`** (mainnet) account?

lewis-sqa: I think this documentation might help: https://github.com/near/near-indexer-for-explorer/#shared-public-access

amirsaran3: Now I am doing it like this so it works for testnet and mainnet.
```js
const response = await fetch(
  `${network.helperUrl}/publicKey/ed25519:${publicKey}/accounts`
);
```

lewis-sqa (approved):

PR #265, opened 2022-04-20
refactor: Created utils package

amirsaran3:
# Description
Created a `utils` package and moved all common utility methods into it.

# Checklist:

- [x] REFACTOR - a PR of this type introduces refactoring.

# Breaking changes


packages/wallet-utils/src/lib/wallet-utils.ts
```diff
@@ -0,0 +1,125 @@
+import { utils, transactions as nearTransactions, Signer } from "near-api-js";
+import {
+  Transaction,
+  Action,
+  ProviderService,
```
lewis-sqa, line 5: To avoid exposing this internal service, we should consider using the JsonProvider from `near-api-js` directly

packages/wallet-utils/src/lib/wallet-utils.ts
```diff
@@ -0,0 +1,125 @@
...
+        return nearTransactions.deleteAccount(beneficiaryId);
+      }
+      default:
+        throw new Error("Invalid action type");
+    }
+  });
+};
+
+const signTransactions = async (
+  transactions: Array<Optional<Transaction, "signerId">>,
```
lewis-sqa, line 87: The defaulting of the `signerId` (`accountId`) shouldn't live here. We should only accept `Array<Transaction>`. This will have an implication on the `viewAccessKey` call and remove the need for the `accountId` parameter.

The reason we ignored this parameter in some wallets is because they only supported a single account, but this can't be the case for a generic method like this.

packages/utils/src/lib/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+interface CreateTransactionParams {
+  accountId: string;
+  publicKey: string;
+  receiverId: string;
+  nonce: number;
+  actions: Array<Action>;
+  hash: string;
+}
+
+const createTransaction = ({
```
lewis-sqa, line 15: I'm not sure this method is terribly useful in practice. I think it would only make sense if it resolved the `nonce` and `hash` but that's what `signTransactions` does.

Maybe out of scope for this task, but we could investigate if NEAR Walle could just use `signTransactions` instead. I'd just suggest reverting the changes in `near-wallet` to just use the `createTransaction` directly from `near-api-js`

packages/wallet-utils/README.md
```diff
@@ -0,0 +1,10 @@
+# @near-wallet-selector/wallet-utils
+
+This is the [Wallet Utils](https://chrome.google.com/webstore/detail/wallet-utils/epapihdplajcdnnkdeiahlgigofloibg) package for NEAR Wallet Selector.
```
lewis-sqa, line 3: This link is wrong here

packages/wallet-utils/src/lib/wallet-utils.ts
amirsaran3: Does this mean that in `/packages/core/src/lib/wallet/wallet.types.ts` we will change `transactions` to `Array<Transaction>`. 
```js
export interface SignAndSendTransactionsParams {
  transactions: Array<Optional<Transaction, "signerId">>;
}
```

Now I can't pass in the param into `signTransactions` because of the type. near-wallet, sender and wallet-connect still need the old param type of `SignAndSendTransactionsParams `.

 How can I solve this?

lewis-sqa: `SignAndSendTransactionsParams` will stay as it is. We'll need to perform the defaulting logic within each wallet implementation which is usually the first account

packages/wallet-utils/src/lib/sign-transactions.ts
```diff
@@ -0,0 +1,54 @@
...
+    }),
+  ]);
+
+  const signedTransactions: Array<nearTransactions.SignedTransaction> = [];
+
+  for (let i = 0; i < transactions.length; i++) {
+    const actions = createAction(transactions[i].actions);
+
+    const transaction = nearTransactions.createTransaction(
+      accountId,
```
lewis-sqa, line 36: We should use `signerId`, remove `accountId` as a parameter and remove the `Optional` type on the `transactions` parameter

packages/wallet-utils/src/lib/create-action.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  const { receiverId, methodNames = [] } = permission;
+  const allowance = permission.allowance
+    ? new BN(permission.allowance)
+    : undefined;
+
+  return transactions.functionCallAccessKey(receiverId, methodNames, allowance);
+};
+
+export const createAction = (actions: Array<Action>) => {
```
lewis-sqa, line 18: I think there's been some confusion on this. The type for this method should be:

```ts
createAction(action: Action): transactions.Action;
```

The idea is the consumer of this function will use `.map` if needed

packages/ledger/src/lib/ledger.ts
```diff
@@ -272,15 +263,19 @@ const Ledger: WalletBehaviourFactory<HardwareWallet> = async ({
...
-          receiverId,
-          actions,
-        },
-      ]);
+      const signedTransactions = await signTransactions(
+        [
+          {
+            receiverId,
+            actions,
+            signerId: signerId!,
```
lewis-sqa, line 271: We'll need to default this to the first account if `signerId` is `undefined`

packages/math-wallet/src/lib/math-wallet.ts
```diff
@@ -104,39 +125,19 @@ const MathWallet: WalletBehaviourFactory<InjectedWallet> = async ({
...
-      );
-
-      const [hash, signedTx] = await nearTransactions.signTransaction(
-        transaction,
+      const signedTransactions = await signTransactions(
+        [
+          {
+            receiverId,
+            actions,
+            signerId: signerId!,
```
lewis-sqa, line 133: Similar case here where we should default the `signerId`

lewis-sqa: You can technically reuse `transformTransactions`:
```ts
transformTransactions([{ signerId, receiverId, actions }])
```

PR #305, opened 2022-05-26
test: Sqac 150 ledger wallet unit testing

amirsaran3:
# Description
Fixes `ledger.spec.ts` tests.

# Checklist:

- [x] TEST - a PR of this type adds more tests.

# Breaking changes

### DIFF
--- a/packages/core/src/lib/testUtils.ts
+++ b/packages/core/src/lib/testUtils.ts
@@ -39,10 +39,7 @@ export const mockWallet = async <Variation extends Wallet>(
   const storeMock = mock<Store>({
     getState: jest.fn(() => {
       return {
-        contract: {
-          contractId: "guest-book.testnet",
-          methodNames: [],
-        },
+        contract: null,
         modules: [],
         accounts: [],
         selectedWalletId: null,
--- a/packages/core/src/lib/testUtils.ts
+++ b/packages/core/src/lib/testUtils.ts
@@ -45,7 +45,7 @@ export const mockWallet = async <Variation extends Wallet>(
         },
         modules: [],
         accounts: [],
-        selectedWalletId: "ledger",
+        selectedWalletId: null,
       };
     }),
   });
--- a/packages/core/src/lib/testUtils.ts
+++ b/packages/core/src/lib/testUtils.ts
@@ -36,12 +36,26 @@ export const mockWallet = async <Variation extends Wallet>(
     return null;
   }
 
+  const storeMock = mock<Store>({
+    getState: jest.fn(() => {
+      return {
+        contract: {
+          contractId: "guest-book.testnet",
+          methodNames: [],
+        },
+        modules: [],
+        accounts: [],
+        selectedWalletId: "ledger",
+      };
+    }),
+  });
+
   const wallet = await module.init({
     id: module.id,
     type: module.type,
     metadata: module.metadata,
     options,
-    store: deps.store || mock<Store>(),
+    store: deps.store || storeMock,
     provider: deps.provider || mock<ProviderService>(),
     emitter: deps.emitter || mock<EventEmitterService<WalletEvents>>(),
     logger: deps.logger || mock<LoggerService>(),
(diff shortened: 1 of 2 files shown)


packages/ledger/src/lib/ledger.spec.ts
```diff
@@ -57,13 +61,32 @@ const createLedgerWallet = async (deps: MockWalletDependencies = {}) => {
...
+  });
+
   provider.viewAccessKey.mockResolvedValue({
     nonce: 0,
     permission: "FullAccess",
     block_height: 0,
     block_hash: "block_hash",
   });
 
+  provider.sendTransaction.mockResolvedValue(getSendTransactionResponse());
```
lewis-sqa, line 88: I'm not sure we need to mock this method as it'll already be mocked by `jest-mock-extended` for the purpose of what we need as we don't really need to assert the return value in our tests as TypeScript helps us enforce this

packages/ledger/src/lib/ledger.spec.ts
```diff
@@ -78,51 +101,101 @@ afterEach(() => {
...
+        signerId: "amirsaran.testnet",
+        receiverId: "guest-book.testnet",
+        actions: [],
+      },
+    ];
+    const result = await wallet.signAndSendTransactions({
+      transactions: transactions,
+    });
+    expect(ledgerClient.sign).toHaveBeenCalled();
+    expect(result).toEqual([getSendTransactionResponse()]);
```
lewis-sqa, line 174: As mentioned in a previous comment, we can omit this assertion

packages/ledger/src/lib/ledger.spec.ts
```diff
@@ -78,51 +101,101 @@ afterEach(() => {
...
+      {
+        signerId: "amirsaran.testnet",
+        receiverId: "guest-book.testnet",
+        actions: [],
+      },
+    ];
+    const result = await wallet.signAndSendTransactions({
+      transactions,
+    });
+    expect(ledgerClient.sign).toHaveBeenCalled();
```
lewis-sqa, line 197: We could improve this slightly by checking it's called `transactions.length` times. Similarly, we could assert `result` returns the same length too (without asserting the values since TypeScript has us covered for this)

packages/ledger/src/lib/ledger.spec.ts
```diff
@@ -1,37 +1,42 @@
...
+import { HardwareWallet, Transaction } from "../../../core/src/lib/wallet";
+import {
   ProviderService,
   JsonStorageService,
 } from "../../../core/src/lib/services";
 import { LedgerClient } from "./ledger-client";
+import {
+  ReadOnlyStore,
+  WalletSelectorState,
+} from "packages/core/src/lib/store.types";
```
lewis-sqa, line 17: This import should be `../../../core/src/lib/store.types`.
We should also use `import type` when import types as we were previously in this file

packages/ledger/src/lib/ledger.spec.ts
```diff
@@ -75,56 +97,107 @@ afterEach(() => {
...
+        signerId: "amirsaran.testnet",
+        receiverId: "guest-book.testnet",
+        actions: [],
+      },
+    ];
+    const result = await wallet.signAndSendTransactions({
+      transactions: transactions,
+    });
+    expect(ledgerClient.sign).toHaveBeenCalled();
+    expect(result.length).toEqual(1);
```
lewis-sqa, line 176: We should reference `transactions.length` to show the link and avoid issues with these tests if the input data was changed to maybe 2 transactions

packages/core/src/lib/testUtils.ts
```diff
@@ -45,7 +45,7 @@ export const mockWallet = async <Variation extends Wallet>(
         },
         modules: [],
         accounts: [],
-        selectedWalletId: "ledger",
+        selectedWalletId: null,
```
lewis-sqa, line 48: We should make `contract` null too

amirsaran3: I am getting an error (Error: Wallet not signed in) when I set it as null

lewis-sqa (approved):

PR #314, opened 2022-06-01
feat: Add Nightly Wallet to Selector

NorbertBodziony:
# Description

This PR adds support for Nightly extension wallet. It works with Solana and Near and has hardware wallet support.
I have added small changes to react template like eager connect for nightly and I think there is bug when wallet is already selected so also fixed that.

Closes # (issue)

# Checklist:


wojciech-cichocki (dismissed):
LGTM

packages/modal-ui/src/lib/components/WalletOptions.tsx
```diff
@@ -76,13 +76,17 @@ export const WalletOptions: React.FC<WalletOptionsProps> = ({
...
             result.push(
               <li
                 key={module.id}
                 id={module.id}
                 className={selected ? "selected-wallet" : ""}
-                onClick={selected ? undefined : handleWalletClick(module)}
+                onClick={
+                  selected
+                    ? // This kinda broke for me if we dont have eager connect
+                      handleWalletClick(module)
```
kujtimprenkuSQA, line 87: - The way we have it it's meant to avoid calling signIn on a wallet that's already signedIn.
Can you please provide more info what kind of problem you had with this code?

examples/react/src/components/Content.tsx
```diff
@@ -24,6 +25,18 @@ const Content: React.FC = () => {
   const [messages, setMessages] = useState<Array<Message>>([]);
   const [loading, setLoading] = useState<boolean>(false);
 
+  // Try eager connect on app launch
+  useEffect(() => {
+    const eagerConnect = async () => {
```
kujtimprenkuSQA, line 30: Why is this code needed?

If you already are `signedIn` with a wallet and the `selectedWalletId` has a value we check that internally when we setup wallet modules.

packages/nightly/src/lib/nightly.ts
```diff
@@ -0,0 +1,170 @@
...
+}
+
+interface NightlyState {
+  wallet: NearNightly;
+}
+
+const setupNightlyState = (meta: InjectedWalletMetadata): NightlyState => {
+  const wallet = window.nightly?.near;
+  if (!wallet) {
+    window.location.href = meta.downloadUrl;
```
kujtimprenkuSQA, line 29: Is there a way to download this wallet extension so we can try it out?

kujtimprenkuSQA (commented):
Please update the main readme to include `Nightly`  wallet too.

examples/react/src/components/Content.tsx
NorbertBodziony: Nightly and I expect other wallet need some kind of session. You need to call `signIn()` or wallet will not able to receive informations from webpage. This is how injected wallets should work IMO.
Nightly knows if you already got permission to connect so popup will not show again.

packages/nightly/src/lib/nightly.ts
NorbertBodziony: YES I left a link in readme. Please dont share for now 😈

NorbertBodziony:
I have moved eager connect to an internal part of wallet, so front end users will not have to do anything.

packages/nightly/src/lib/nightly.ts
lewis-sqa: To align with other injected wallets, we need to avoid automatically redirecting like this. Wallet Selector will only display wallets that are currently installed. `downloadUrl` will likely be used in the future to handle this use case.

Take a look at the `isInstalled` logic for Sender and Math Wallet to return `null`

packages/nightly/src/lib/nightly.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    async signOut() {
+      await _state.wallet.disconnect();
+    },
+
+    async getAccounts() {
+      const nearAccount: Account = {
+        accountId: _state.wallet.account.accountId,
+      };
+      return [nearAccount];
```
lewis-sqa, line 75: We should return an empty array if there's no account (I think Nightly returns an empty string for `accountId` and a public key with all 1s) and use this method in `signIn` to avoid duplicate checking logic

packages/nightly/src/lib/nightly.ts
```diff
@@ -0,0 +1,175 @@
...
+        new PublicKey(_state.wallet.account.publicKey),
+        receiverId || contract!.contractId,
+        // @ts-expect-error
+        ++blockInfo.nonce,
+        actions.map((a) => createAction(a)),
+        blockHash
+      );
+      const signedTransactions = await _state.wallet.signTransaction(tx);
+      const result = await provider.sendTransaction(signedTransactions);
+      return result;
```
lewis-sqa, line 103: We should consider using `signTransactions` from our `wallet-utils` package to clean up and standardise this (and `signAndSendTransactions`) logic. See Math Wallet and Ledger implementations - the main change will be implementing a `Signer` interface

examples/react/src/components/Content.tsx
NorbertBodziony: Yes if wallet is locked this will trigger popup to unlock. Why is that an issue ? This is common behaviour.

lewis-sqa: This flow doesn't occur with the other two injected wallets (Sender and Math Wallet). It's an issue because a user will navigate to a dApp they previously signed in to using Wallet Selector and immediately see the unlock prompt

NorbertBodziony: I am not focusing API on near alone. I am trying to fit into all blockchains (Solana/Near now) and on Solana this is common.
If you guys insist I will create a method that connects only if user already approved app and wallet is unlocked.

packages/nightly/src/lib/nightly.ts
NorbertBodziony: I don't like implementing Signer. Our API handles only `Transactions`. Implementing `Signer` API will force me to deserialize `Uint8Array` into `Transaction` again also we return already `SignedTransaction` so we will need to cut Signature from it. A lot of changes just to fit into `Signer` schema. It's possible but from my point it will create more code.

packages/nightly/src/lib/nightly.ts
NorbertBodziony: Refactored so extension is still visible but will redirect only on click.

packages/nightly/src/lib/nightly.ts
NorbertBodziony: Fixed

packages/nightly/src/lib/nightly.ts
```diff
@@ -0,0 +1,175 @@
...
+  logger,
+  metadata,
+}) => {
+  const _state = setupNightlyState(metadata);
+  const provider = new JsonRpcProvider({ url: options.network.nodeUrl });
+  const state = store.getState();
+
+  if (state.selectedWalletId === "nightly") {
+    await _state.wallet.connect();
+  }
```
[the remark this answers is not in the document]
NorbertBodziony: Fixed but will require update of chrome extension.

NorbertBodziony:
Fixed most of your comments. Eager connect requires update of extension it will take a couple of days cuz of Chrome.
Right now everything in terms of behaviour should be in line with other extensions like Sender. The only difference is that we show extension even if the user does not have this installed.

lewis-sqa:
> We use implicit addresses for now we are going to work on custom ones in later releases. Demo only works if you already have some Near on your account since without it, you can't query it.

Not sure I fully understand the last part of this. When I import my existing NEAR account that has some NEAR on it, it resolves to an incorrect account id. From what I understand, Nightly is correctly resolving my seed phrase in terms of the key pair, but deriving an implicit account id instead of linking to my `.testnet` account.

I've since created a new account directly through Nightly and sent some NEAR to it so I can test this extension. Are there plans to resolve this issue soon as I can foresee a lot of users having issues importing just like our team has?

packages/nightly/src/lib/nightly.ts
lewis-sqa: This isn't a concern that should be handled by the wallet implementations but instead the UI such as `modal-ui`. As mentioned in my previous comment, the wallet should return `null` if it's not installed. The benefit here is we can be sure the wallet exists without introducing lots of `?.` checks.

They'll be some work in the future to address this particular issue in `modal-ui` where the `metadata.downloadUrl` will be used to redirect but this behaviour shouldn't be part of the wallet.

packages/modal-ui/src/lib/components/WalletOptions.tsx
lewis-sqa: This change should be reverted to address the point made by @kujtimprenkuSQA

packages/nightly/src/lib/injected-nightly.ts
```diff
@@ -0,0 +1,46 @@
...
+    eagerConnect?: boolean
+  ): Promise<NearAccount>;
+  disconnect(): Promise<void>;
+  signTransaction: (
+    transaction: NearTransaction
+  ) => Promise<NearSignedTransaction>;
+  signAllTransactions: (
+    transaction: Array<NearTransaction>
+  ) => Promise<Array<NearSignedTransaction>>;
+}
```
lewis-sqa, line 46: We should be able to simply define interfaces in this file instead of declaring classes:

```suggestion
export NightlyAccount {
  accountId: string;
  publicKey: NearPublicKey;
}

export interface NearNightly {
  account: NightlyAccount;
  connected: boolean;
  signTransaction: (transaction: NearTransaction) => Promise<NearSignedTransaction>;
  signAllTransactions: (transaction: Array<NearTransaction>) => Promise<Array<NearSignedTransaction>>;
  connect: (onDisconnect?: () => void, eagerConnect?: boolean) => Promise<NightlyAccount>;
  disconnect: () => Promise<void>;
}

export interface InjectedNightly {
  near: NearNightly;
  invalidate: () => void;
}
```

packages/nightly/src/lib/nightly.ts
```diff
@@ -0,0 +1,175 @@
...
+  const _state = setupNightlyState();
+  const currentState = store.getState();
+  if (currentState.selectedWalletId === "nightly") {
+    try {
+      // eager connect to the wallet
+      await _state?.connect(undefined, true);
+    } catch {
+      // ignore
+    }
+  }
```
lewis-sqa, line 40: As mentioned previously, once we've addressed the installed status of the extension, we can assume `window.nightly.near` exists which will clean up the defensive approach that's be used (e.g. `?.`).

```suggestion
declare global {
  interface Window {
    nightly: InjectedNightly | undefined;
  }
}

interface NightlyState {
  wallet: InjectedNightly;
}

const setupNightlyState = async (store: WalletSelectorStore): NightlyState => {
  const { selectedWalletId } = store.getState();
  const wallet = window.nightly.near!
  
  // Attempt to reconnect wallet if previously selected.
  if (selectedWalletId === "nightly") {
      await wallet.connect(undefined, true).catch(() => null);
  }

  return {
    wallet,
  };
};

const Nightly: WalletBehaviourFactory<InjectedWallet> = async ({
  metadata,
  store,
  logger,
  provider,
}) => {
  const _state = await setupNightlyState(store);
```

NorbertBodziony:
> I've since created a new account directly through Nightly and sent some NEAR to it so I can test this extension. Are there plans to resolve this issue soon as I can foresee a lot of users having issues importing just like our team has?

Yes we are going to work on this, but we are still working on best possible UX right now we just use implicit address.
It also does not require you to deposit any near and is much easier for folks coming from ETH/SOL

packages/nightly/src/lib/injected-nightly.ts
NorbertBodziony: Fixed https://github.com/near/wallet-selector/pull/314/commits/55b918ebf6991e08b32220b4197d9443c6ae760d

packages/nightly/src/lib/nightly.ts
NorbertBodziony: refactor here https://github.com/near/wallet-selector/pull/314/commits/33bb3202b47ffc3f0fe2dbc2fc8460ab56816baf

packages/nightly/src/lib/nightly.ts
```diff
@@ -0,0 +1,194 @@
...
+    return [nearAccount];
+  };
+  return {
+    // nightly does not support delegating signing right now
+    async signIn() {
+      // If wallet does not exist user will be redirected to download page
+      if (!_state) {
+        window.location.href = metadata.downloadUrl;
+        throw new Error("Redirecting to download");
+      }
```
lewis-sqa, line 75: This should be removed to align with other wallets.

packages/nightly/src/lib/nightly.ts
```diff
@@ -0,0 +1,194 @@
...
+      const existingAccount = _state.wallet.account.accountId;
+
+      if (existingAccount) {
+        const nearAccount: Account = {
+          accountId: _state.wallet.account.accountId,
+        };
+        return [nearAccount];
+      }
```
lewis-sqa, line 83: We should reuse `getAccounts` here

lewis-sqa (commented):
Last few comments from me. There's also the change in `WalletOptions.tsx` that needs reverting

packages/nightly/src/lib/nightly.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  const currentState = store.getState();
+  if (currentState.selectedWalletId === "nightly") {
+    try {
+      // eager connect to the wallet
+      await _state.wallet.connect(undefined, true);
+    } catch {
+      // ignore
+    }
+  }
```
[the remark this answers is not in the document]
NorbertBodziony: fix https://github.com/near/wallet-selector/pull/314/commits/45125a100f0dc1944e462a92180be28072cea983

packages/nightly/src/lib/nightly.ts
NorbertBodziony: fix https://github.com/near/wallet-selector/pull/314/commits/45125a100f0dc1944e462a92180be28072cea983

packages/nightly/src/lib/nightly.ts
NorbertBodziony: fix https://github.com/near/wallet-selector/pull/314/commits/45125a100f0dc1944e462a92180be28072cea983

PR #328, opened 2022-06-17
feat: Added active property to AccountState

amirsaran3:
# Description

Closes # (issue)

# Checklist:

# Breaking changes


packages/core/src/lib/store.ts
```diff
@@ -62,9 +76,16 @@ const reducer = (
         return state;
       }
 
+      const accountStates = accounts.map((account, i) => {
+        return {
+          ...account,
+          active: i === 0,
```
lewis-sqa, line 82: This scenario is a little more involved as we should try to maintain the existing account that's selected unless it's been removed (in which case we would default to the first one)

packages/ledger/src/lib/ledger.ts
```diff
@@ -1,15 +1,15 @@
 import { isMobile } from "is-mobile";
 import { TypedError } from "near-api-js/lib/utils/errors";
 import { signTransactions } from "@near-wallet-selector/wallet-utils";
-import type {
```
lewis-sqa, line 4: We should maintain this import style here but have a new import statement for `getActiveAccount`

lewis-sqa: Same for the other wallets

lewis-sqa (commented):
Looking good 👍

I think we're just missing the functionality for changing the active account. Something like `selector.setActiveAccount(accountId: string)`

packages/core/src/lib/helpers/getActiveAccount.ts
```diff
@@ -0,0 +1,8 @@
+import { AccountState } from "../store.types";
+import { WalletSelectorStore } from "../wallet-selector.types";
+
+export const getActiveAccount = (
+  store: WalletSelectorStore
+): AccountState | null => {
+  return store.getState().accounts.find((account) => account.active) || null;
+};
```
lewis-sqa, line 8: Maybe we could make this reusable in the dApp by switching this to use `WalletSelectorState` instead of the `WalletSelectorStore`

packages/core/src/lib/wallet-selector.ts
```diff
@@ -48,6 +48,12 @@ export const setupWalletSelector = async (
 
         return wallet;
       },
+      setActiveAccount: (accountId: string) => {
+        store.getState().accounts = store.getState().accounts.map((account) => {
+          account.active = account.accountId === accountId;
+          return account;
+        });
```
lewis-sqa, line 55: To update the state, we must dispatch a new action so it can go through the store reducer. It's important that this happens instead of mutating like this as it allows subscribes to be notified

packages/core/src/lib/services/wallet-modules/wallet-modules.service.ts
```diff
@@ -189,6 +189,17 @@ export class WalletModules {
...
+    emitter.on("setActiveAccount", async ({ accounts, accountId }) => {
+      if (!accounts.length) {
+        return this.signOutWallet(module.id);
+      }
+
+      this.store.dispatch({
+        type: "SET_ACTIVE_ACCOUNT",
+        payload: { accountId, accounts },
+      });
+    });
```
lewis-sqa, line 201: This emitter is really for events that occur within a wallet. We should move back to the `setActiveAccount` method on the `selector` object. We should only need the `accountId` as the store already contains the list of accounts.

I'm a little confused by the sign out logic we have here too. What's the reason for that?

lewis-sqa (commented):
Changes outstanding are:
- Handling in `ACCOUNTS_CHANGED` when the active account no longer be applicable because it's been removed. It should default to the first account.
- `SET_ACTIVE_ACCOUNT` doesn't require `accounts` as this can be referenced in the store. Ideally we should do some validation in `setActiveAccount` on the selector to ensure the accountId is actually valid as we'll want to avoid a case where there's no active account but we have a list of accounts.

packages/core/src/lib/store.ts
```diff
@@ -19,10 +19,21 @@ const reducer = (
     case "SETUP_WALLET_MODULES": {
       const { modules, accounts, contract, selectedWalletId } = action.payload;
 
+      const activeAccountIndex = state.accounts.findIndex(
```
lewis-sqa, line 22: We don't need this logic for `SETUP_WALLET_MODULES` as we won't have an active account. We can always assume the first account, unless we decided to persist the active account in the future

packages/core/src/lib/wallet/wallet.types.ts
```diff
@@ -67,6 +67,7 @@ export type WalletEvents = {
   };
   signedOut: null;
   accountsChanged: { accounts: Array<Account> };
+  setActiveAccount: { accountId: string; accounts: Array<AccountState> };
```
lewis-sqa, line 70: We can remove this event now

lewis-sqa (approved):

examples/react/src/contexts/WalletSelectorContext.tsx
```diff
@@ -85,7 +59,17 @@ export const WalletSelectorContextProvider: React.FC = ({ children }) => {
...
+    if (!state.accounts.length) {
+      setAccounts([]);
+    }
+
+    const newAccount = state.accounts.find((account) => account.active);
+
+    if (newAccount) {
+      _selector.setActiveAccount(newAccount.accountId);
+      setAccounts(state.accounts);
+    }
```
lewis-sqa, line 72: We should be able to replace this (and in the subscription further down) with simply:

```suggestion
setAccounts(state.accounts);
```

examples/angular/src/app/app.component.ts
```diff
@@ -87,7 +63,17 @@ export class AppComponent implements OnInit {
...
+      this.accountId = null;
+      this.accounts = [];
+    }
+
+    const newAccount = state.accounts.find((account) => account.active);
+
+    if (newAccount) {
+      this.accountId = newAccount.accountId;
+      this.accounts = state.accounts;
+    }
```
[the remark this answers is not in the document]
lewis-sqa: Regarding the comment, in Angular, this will be a problem:
```ts
if (prevAccountId !== this.accountId) {
  this.getAccount().then((account) => {
    this.account = account;
  });
}
```

examples/react/src/contexts/WalletSelectorContext.tsx
```diff
@@ -114,24 +89,37 @@ export const WalletSelectorContextProvider: React.FC = ({ children }) => {
...
+
+          return;
+        }
+
+        const newAccount = nextAccounts.find((account) => account.active);
+
+        if (newAccount) {
+          selector.setActiveAccount(newAccount.accountId);
+          setAccounts(nextAccounts);
+        }
```
lewis-sqa, line 103: This can all be replaced with `setAccounts(nextAccounts)`

lewis-sqa (approved):

PR #331, opened 2022-06-21
fix: Sender Icon

keeplearning-near:
# Description

The sender wallet website has been updated with a new icon. Accordingly, the wallet-selector should be updated at the same time.


keeplearning-near:
Preview image: 
image

lewis-sqa (approved):
@keeplearning-near thanks for this. We'll look to get this into the next release 👍

PR #339, opened 2022-06-29
feat: Show `Wallet Not Installed` View for Injected Wallets

kujtimprenkuSQA:
# Description

This PR addresses the case when an injected wallet is supported by a dApp but the user has not installed that specific wallet in their browser.

- Clicking on a wallet option that is not installed shows the `Wallet Not Installed` view

Closes # (issue)

# Checklist:

# Breaking changes


packages/math-wallet/src/lib/math-wallet.ts
```diff
@@ -155,6 +155,7 @@ export const setupMathWallet = ({
         downloadUrl:
           "https://chrome.google.com/webstore/detail/math-wallet/afbcbjpbpfadlkmhmclhkeeodmamcflc",
         deprecated: false,
+        available: installed,
```
lewis-sqa, line 158: We'll need to modify the `setupInstance` method in `wallet-modules` to avoid calling `init` when a wallet is unavailable. We run the risk of causing crashes otherwise 🙁

packages/modal-ui/src/lib/components/WalletNotInstalled.tsx
```diff
@@ -1,25 +1,29 @@
...
-import { Wallet } from "@near-wallet-selector/core";
+import { ModuleState } from "@near-wallet-selector/core";
 
 interface WalletNotInstalledProps {
-  wallet: Wallet;
+  module: ModuleState & {
+    metadata: {
+      downloadUrl?: string;
+    };
+  };
```
lewis-sqa, line 9: We shouldn't need to do this, but it looks like I didn't quite get the types right with `ModuleState` 🙁  This is fine for now

kujtimprenkuSQA: Yeah the `downloadUrl` is not being added to the metadata of `ModuleState` properly to keep typescript "happy" I added this kind of workaround.

PR #355, opened 2022-07-18
feat: Add Tokenary wallet

grachyov:
# Description

Add Tokenary Safari extension support.

Tokenary implements an API almost identical to Sender wallet.

For example, it works with https://app.burrow.cash in Safari when you enable Tokenary extension and select Sender wallet from wallet selector.

# Checklist:

# Breaking changes


lewis-sqa:
Hey @grachyov,

I've done some testing with Tokenary this morning and found a few problems with it's current implementation:

- Imported wallets with named account ids (`test.testnet`) aren’t working as it assumes implicit account ids.
- I can’t seem to change the network to `testnet`.
- `signAndSendTransaction` resolved even when there’s an error. What’s the reason it doesn’t reject instead?
- It declares the `isSender` flag as well as `isTokenary` which will confuse dApps.
- `isSignedIn` returns `true` before I’ve interacted with the dApp.
- It populates the `accountId` with my imported account before I’ve signed in.
- `signOut` doesn’t seem to clear the `accountId` or `isSignedIn` status.

grachyov:
Hey @lewis-sqa, thanks for the detailed feedback!

* Is there only support for [FunctionCall](https://nomicon.io/RuntimeSpec/Actions#functioncallaction) Actions? — Yes, only `FunctionCall` actions are supported right now.

* Are the error messages exactly the same? — No. Are dapps relying on these?

* Are all the events applicable? — Do you mean emitting `signIn`, `signOut`, `accountChanged` and `rpcChanged`?

* How is sign in and sign out handled? Does it delete the key upon sign out? — I assumed from [this line](https://github.com/near/wallet-selector/blob/e9e2f8ed027ce6764491dc9d4cc4c4e6119a8841/packages/sender/src/lib/sender.ts#L156) and from using dapps, that responding with `true` without actually creating an `AccessKey` is enough right now. Should we create it anyways?

* Imported wallets with named account ids (test.testnet) aren’t working as it assumes implicit account ids. — Would it be necessary for the initial integration? If I import an account from Trust wallet, I would actually expect an implicit account id since it is displayed this way in Trust wallet.

* I can’t seem to change the network to testnet. — Tokenary does not support testnet yet. Should we add it?

* signAndSendTransaction resolves even when there’s an error. What’s the reason it doesn’t reject instead? — I assumed that [this is how Sender handles an error](https://github.com/near/wallet-selector/blob/e9e2f8ed027ce6764491dc9d4cc4c4e6119a8841/packages/sender/src/lib/sender.ts#L216) and we should do it the same way. Would reject be better here?

* It declares the isSender flag as well as isTokenary which will confuse dApps. — We've added it to be able to connect Tokenary by selecting Sender option while integration is not completed. Going to remove isSender flag when the integration is ready to be merged.

* isSignedIn returns true before I’ve interacted with the dApp. It populates the accountId with my imported account before I’ve signed in. — As I mentioned, we do not actually create `AccessKey` and just return `accountId` and `true` for `isSignedIn` right after user selects an account in a popup. Should we fix this behaviour?

* signOut doesn’t seem to clear the accountId or isSignedIn status. — Will fix it.

lewis-sqa:
@grachyov, apologies for the delayed response:

> Are the error messages exactly the same? — No. Are dapps relying on these?

Sadly, Wallet Selector depends on the messages to identify cancel scenarios (e.g. Sender):
https://github.com/near/wallet-selector/blob/main/packages/sender/src/lib/sender.ts#L73

> Are all the events applicable? — Do you mean emitting signIn, signOut, accountChanged and rpcChanged?

Yeah that's right. Does Tokenary support all of these and with the same event payloads? You mentioned Tokenary only supports `mainnet` so I'm guessing `rpcChanged` isn't applicable?

> How is sign in and sign out handled? Does it delete the key upon sign out? — I assumed from [this line](https://github.com/near/wallet-selector/blob/e9e2f8ed027ce6764491dc9d4cc4c4e6119a8841/packages/sender/src/lib/sender.ts#L156) and from using dapps, that responding with true without actually creating an AccessKey is enough right now. Should we create it anyways?

Until a [standard officially exists](https://github.com/near/NEPs/pull/370) you're right, though "sign in" generally means creating a `FunctionCall` access key for a dApp's smart contract in NEAR which is why it accepts a `contractId` and (optionally) `methodNames`. There are cases where these limited access keys aren't applicable such as dApps that don't make any gas-only function call transactions (but again a standard such as the one referenced will eventually address this problem).

> Imported wallets with named account ids (test.testnet) aren’t working as it assumes implicit account ids. — Would it be necessary for the initial integration? If I import an account from Trust wallet, I would actually expect an implicit account id since it is displayed this way in Trust wallet.

I'm not an expert on this recent change in NEAR but the current integration doesn't work when a user has a named account. For instance, Tokenary thinks my `lewis-sqa-2.testnet` account is `80d028da4035d532519c081d71901b19e4ded4f3eb0713c37fad2bd0f2d5c7ec`.

> I can’t seem to change the network to testnet. — Tokenary does not support testnet yet. Should we add it?

If you could add support for this network, it would greatly help us test the integration 🙂 

> signAndSendTransaction resolves even when there’s an error. What’s the reason it doesn’t reject instead? — I assumed that [this is how Sender handles an error](https://github.com/near/wallet-selector/blob/e9e2f8ed027ce6764491dc9d4cc4c4e6119a8841/packages/sender/src/lib/sender.ts#L216) and we should do it the same way. Would reject be better here?

You're right, though (in my honest opinion) I don't think Sender is a great example of good practices with JavaScript, particularly in this case where a core concept of Promises is effectively ignored 🙁 

> It declares the isSender flag as well as isTokenary which will confuse dApps. — We've added it to be able to connect Tokenary by selecting Sender option while integration is not completed. Going to remove isSender flag when the integration is ready to be merged.

Ah, that makes sense. No problem 👍 

> isSignedIn returns true before I’ve interacted with the dApp. It populates the accountId with my imported account before I’ve signed in. — As I mentioned, we do not actually create AccessKey and just return accountId and true for isSignedIn right after user selects an account in a popup. Should we fix this behaviour?

The problem I had was I don't recall selected an account (maybe because I only imported one?), but the object had a reference to the account I imported. Basing behaviour on Math Wallet (which is probably more similar to the approach you're heading towards in terms of no `FunctionCall` access key support), I'd expect these fields to be populated only once I've called `signIn`

> signOut doesn’t seem to clear the accountId or isSignedIn status. — Will fix it.

Excellent 🎉

grachyov:
@lewis-sqa, thanks for your response!
I'll add changes to Tokenary and be back soon.

AmmarHumackicSQA:
Hey @grachyov , when can we expect these changes as we planned it for our next release? Thanks!

grachyov:
Hey @AmmarHumackicSQA , I am not quite sure yet.

grachyov:
Hey @janewang , this one or ivan@tokenary.io is good.

AmmarHumackicSQA:
Hey @grachyov , can you update us on the progress regarding the mentioned changes and confirm that the [criteria](https://github.com/near/wallet-selector/blob/main/CONTRIBUTING.md#listing-criteria-for-third-party-wallet-on-wallet-selector) are met or can be met within the required timeline? Thanks!

AmmarHumackicSQA:
> Hey @grachyov , can you update us on the progress regarding the mentioned changes and confirm that the [criteria](https://github.com/near/wallet-selector/blob/main/CONTRIBUTING.md#listing-criteria-for-third-party-wallet-on-wallet-selector) are met or can be met within the required timeline? Thanks!

Hi @grachyov any news on this? Thanks.

AmmarHumackicSQA:
> > Hey @grachyov , can you update us on the progress regarding the mentioned changes and confirm that the [criteria](https://github.com/near/wallet-selector/blob/main/CONTRIBUTING.md#listing-criteria-for-third-party-wallet-on-wallet-selector) are met or can be met within the required timeline? Thanks!
> 
> Hi @grachyov any news on this? Thanks.

Hey @grachyov should we leave this PR open, and can you share more info regarding the status of the work you are doing?

AmmarHumackicSQA:
@grachyov gentle bump 🙂

DamirSQA:
Hey @grachyov , any news on this one?

AmmarHumackicSQA:
Hey, @grachyov can you let us know what is the status regarding this PR? We might need to close it soon if we don't get any response. Thanks

PR #364, opened 2022-07-25
fix: Ledger will now show in UI if not available

amirsaran3:
# Description
Ledger will now show in UI if not available. When clicked an error will show up.

# Checklist:

# Breaking changes

### DIFF
--- a/packages/ledger/src/lib/ledger.ts
+++ b/packages/ledger/src/lib/ledger.ts
@@ -276,7 +276,7 @@ export function setupLedger({
     const mobile = isMobile();
     const supported = isLedgerSupported();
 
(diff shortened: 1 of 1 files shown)


lewis-sqa (commented):
Not sure if this will work? It looks like `handleWalletClick` in `modal-ui` only navigates to `WalletNotInstalled` when it's an injected wallet and not available? Even then, that screen uses terminology like "installing" which would be confusing with Ledger

lewis-sqa:
My bad, after some discussion it seems we're relying on the error thrown by the Ledger library to alert the user it's not available 👍

lewis-sqa (approved):

PR #369, opened 2022-07-28
refactor: Support Single Derivation Path with Multiple Account Id-s

kujtimprenkuSQA:
# Description

The reason for this refactoring and switching to a single derivation path with multiple account id(s) is to simplify the UI/UX and apparently this is a common use case in the ecosystem from the feedback we've got in the last calls with NEAR.

- Added support to select a single derivation path.
- Select one or multiple account id(s) for the chosen derivation path.
- Added fallback when indexer service fails (provide account id manually).

Closes # (issue)

# Checklist:

- [x] REFACTOR - a PR of this type introduces refactoring.

# Breaking changes

### DIFF
--- a/packages/modal-ui/src/lib/components/DerivationPath.tsx
+++ b/packages/modal-ui/src/lib/components/DerivationPath.tsx
@@ -94,27 +94,6 @@ export const DerivationPath: React.FC<DerivationPathProps> = ({
     }
   };
 
-  const handleSignIn = () => {
-    const mapAccounts = accounts.map((account: HardwareWalletAccount) => {
-      return {
-        derivationPath: account.derivationPath,
-        publicKey: account.publicKey,
-        accountId: account.accountId,
-      };
-    });
-
-    return hardwareWallet!
-      .signIn({
-        contractId: options.contractId,
-        methodNames: options.methodNames,
-        accounts: mapAccounts,
-      })
-      .then(() => onConnected())
-      .catch((err) => {
-        onError(`Error: ${err.message}`);
-      });
-  };
-
   const handleValidateAccount = async () => {
     const wallet = await selector.wallet(params.walletId);
 
@@ -154,13 +133,6 @@ export const DerivationPath: React.FC<DerivationPathProps> = ({
       setConnecting(false);
     }
   };
-  const handleEnterClick: KeyboardEventHandler<HTMLInputElement> = async (
-    e
-  ) => {
-    if (e.key === "Enter") {
-      await handleValidateAccount();
-    }
-  };
 
   const handleAddCustomAccountId = async () => {
     try {
@@ -185,6 +157,35 @@ export const DerivationPath: React.FC<DerivationPathProps> = ({
       onError(message);
     } finally {
       setConnecting(false);
+    }
+  };
+
+  const handleSignIn = () => {
+    const mapAccounts = accounts.map((account: HardwareWalletAccount) => {
+      return {
+        derivationPath: account.derivationPath,
+        publicKey: account.publicKey,
+        accountId: account.accountId,
+      };
+    });
+
+    return hardwareWallet!
+      .signIn({
+        contractId: options.contractId,
+        methodNames: options.methodNames,
+        accounts: mapAccounts,
+      })
+      .then(() => onConnected())
+      .catch((err) => {
+        onError(`Error: ${err.message}`);
+      });
+  };
+
+  const handleEnterClick: KeyboardEventHandler<HTMLInputElement> = async (
+    e
+  ) => {
+    if (e.key === "Enter") {
+      await handleValidateAccount();
     }
   };
 
--- a/packages/modal-ui/src/lib/components/DerivationPath.tsx
+++ b/packages/modal-ui/src/lib/components/DerivationPath.tsx
@@ -142,7 +142,6 @@ export const DerivationPath: React.FC<DerivationPathProps> = ({
       if (!multipleAccounts) {
         setRoute("OverviewAccounts");
       } else {
-        setConnecting(false);
         setRoute("ChooseAccount");
       }
     } catch (err) {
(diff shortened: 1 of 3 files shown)


packages/modal-ui/src/lib/components/DerivationPath.tsx
```diff
@@ -32,46 +40,23 @@ export const DerivationPath: React.FC<DerivationPathProps> = ({
...
-  };
-
-  const handleDerivationPathRemove = (index: number) => {
-    setDerivationPaths((prevDerivationPaths) => {
-      prevDerivationPaths.splice(index, 1);
-      return [...prevDerivationPaths];
+  const handleAddAccount = (account: HardwareWalletAccountState) => {
+    setAccounts((prevAccounts) => {
+      prevAccounts.push(account);
+      return [...prevAccounts];
```
lewis-sqa, line 55: We should avoid mutative operations such as `.push` using the following:
```ts
return [...prevAccounts, account]
```

packages/modal-ui/src/lib/components/DerivationPath.tsx
```diff
@@ -15,12 +19,16 @@ interface DerivationPathProps {
...
 
-export interface HardwareWalletAccountState {
-  derivationPath: string;
-  publicKey: string;
+export type HardwareWalletAccountState = HardwareWalletAccount & {
   accountIds: Array<string>;
-  selectedAccountId: string;
-}
+  selected: boolean;
+};
```
lewis-sqa, line 25: So accounts for Ledger are no longer grouped by derivationPath. We can simply express them as:
```ts
export interface HardwareWalletAccountState {
  derivationPath: string;
  publicKey: string;
  accountId: string;
}
```
Of course this means we'll see the same `derivationPath` and `publicKey` when multiple linked accounts are found from the indexer which is fine 🙂

packages/modal-ui/src/lib/components/DerivationPath.tsx
```diff
@@ -83,54 +68,55 @@ export const DerivationPath: React.FC<DerivationPathProps> = ({
...
-        const accountIds = await getAccountIdsFromPublicKey(publicKey);
-
-        accounts.push({
-          derivationPath,
-          publicKey,
-          accountIds,
-          selectedAccountId: accountIds[0],
-        });
-      }
+  const resolveAccount = async (
```
lewis-sqa, line 77: Probably just a naming thing but this is really `resolveAccounts` and should maybe return the list of accounts from `getAccountIds`?

lewis-sqa: In fact this abstraction causes problems in `handleValidateAccount`  as we should only display the custom input screen if we get an error from `getAccountIds` and not when the user rejects the `getPublicKey` request. No accounts should display a screen similar to Mutisafe

packages/modal-ui/src/lib/components/DerivationPath.tsx
```diff
@@ -229,71 +205,100 @@ export const DerivationPath: React.FC<DerivationPathProps> = ({
...
-            const mapAccounts = accounts.map((account) => {
-              return {
-                derivationPath: account.derivationPath,
-                publicKey: account.publicKey,
-                accountId: account.selectedAccountId,
-              };
+            setAccounts((prevAccounts) => {
+              prevAccounts = prevAccounts.filter((account) => account.selected);
+
+              return [...prevAccounts];
```
lewis-sqa, line 257: We should avoid reassigning here and simply spread the filter or give it a name like `selectedAccounts`

packages/modal-ui/src/lib/components/HardwareWalletAccountsForm.tsx
```diff
@@ -1,58 +1,66 @@
...
+  accounts,
+  onSelectedChanged,
   onSubmit,
 }) => {
+  const [disableButton, setDisableButton] = useState(false);
+
+  useEffect(() => {
+    const selected = accounts.some((x) => x.selected);
+    setDisableButton(!selected);
+  }, [accounts]);
```
lewis-sqa, line 23: We can remove this state as it can be derived directly from `accounts`
```tsx
disabled={accounts.some((x) => x.selected)}
```

lewis-sqa (commented):
Few comments from me 👍

packages/modal-ui/src/lib/components/DerivationPath.tsx
```diff
@@ -74,48 +67,32 @@ export const DerivationPath: React.FC<DerivationPathProps> = ({
...
+        const selected = i === 0;
+        foundAccounts.push({
+          derivationPath,
+          publicKey,
+          accountId: accountIds[i],
+          selected,
+        });
+      }
+
+      return foundAccounts;
```
lewis-sqa, line 90: We could use `.map` here on `accountIds` instead of a for loop

packages/modal-ui/src/lib/components/DerivationPath.tsx
```diff
@@ -233,6 +232,29 @@ export const DerivationPath: React.FC<DerivationPathProps> = ({
         </div>
       )}
 
+      {route === "NoAccountsFound" && (
+        <div className="no-accounts-found-wrapper">
+          <p>
+            Can't found any account associated with this Ledger. Please create a
+            new NEAR account on{" "}
+            <a href="https://app.mynearwallet.com/create" target="_blank">
```
lewis-sqa, line 240: We'll need to make this dynamic based on the network. Maybe we do need to make `walletUrl` part of the `network` object after all? 😩

kujtimprenkuSQA: Did not add it as part of the `network` mostly because we're trying to connect to Ledger but these url(s) are not related to ledger, maybe in the future we could make this url optional as part of the `setupModal()`.

I made this dynamic based on the network it will be either:

https://testnet.mynearwallet.com/create or
https://app.mynearwallet.com/create

lewis-sqa (approved):

PR #376, opened 2022-08-04
Recommended Wallets Module

kujtimprenkuSQA:
# Description
This package includes these wallets: NEAR Wallet, MyNearWallet, Sender, Ledger, WalletConnect.
In this PR we group the setup of the mentioned wallets above.
- Created new package `@near-wallet-selector/recommended/wallets`
- Exposed `setupRecommendedWallets`.
- Added `whitelist` optional property if dApp wants to include only a set of the recommended wallets.

## Work Remaining
- [ ] Update documentation after finishing implementation.

Closes # (issue)

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings

# Type of change. This type of change is the main reason for the PR.

- [ ] FIX - a PR of this type patches a bug.
- [ ] FEATURE - a PR of this type introduces a new feature.
- [ ] BUILD - a PR of this type introduces build changes.
- [ ] CI - a PR of this type introduces CI changes.
- [ ] DOCS - a PR of this type introduces DOCS improvement.
- [ ] STYLE - a PR of this type introduces style changes.
- [ ] REFACTOR - a PR of this type introduces refactoring.
- [ ] PERFORMANCE - a PR of this type introduces performance changes.
- [ ] TEST - a PR of this type adds more tests.
- [ ] CHORE - a PR introduces other changes than the specified above.

# Breaking changes

- [ ] BREAKING CHANGE - SPECIFY: _______
- [ ] NO BREAKING CHANGE - this PR doesn't contain any breaking changes and it's backwards compatible

PR #379, opened 2022-08-05
feat: Default Wallets Module

kujtimprenkuSQA:
# Description

This package includes these default wallets: 
 - MyNearWallet
 - Ledger

This PR groups the setup of the mentioned wallets above with default options.

Created new package `@near-wallet-selector/default-wallets`
Exposed `setupDefaultWallets()`.

Follow the docs here to learn more on how to use this package:  [Readme]( https://github.com/near/wallet-selector/tree/SQC-154/recommended-wallets-2/packages/default-wallets)

Closes # (issue)

# Checklist:

# Breaking changes


lostpebble:
Hi @kujtimprenkuSQA ,

I'm curious what the benefit of this package is? Consumers of the wallet-selector package can already select which wallets they'd like to implement using the individual packages. I feel like only putting a subset of wallets in this "recommended package" kinda goes against the ethos of the Near ecosystem being "decentralized" and "open", this seems to give a rather biased pathway for folks to implement external wallets.

In the end- the amount of effort required to import and use the individual wallets is almost exactly the same as importing them all from this "recommended-wallets" package. Why not just create a list of "recommended wallets" in the Readme to import, if you'd really like to only recommend the tried and tested wallets? This would be something a lot easier to change later on down the line as you can recommend more wallets.

And if this is the direction you guys are going to head, as a developer in a wallet team myself- what would it take for us to get on this "recommended package" ?

janewang:
@lostpebble Thanks for your comment.  The developer ultimately has complete control over which wallets to show on their dapp. The point we are asking all dapp developers to move away from hardcoding a single wallet url and use wallet-selector as well as onboarding wallets to wallet-selector is very much in the spirit of decentralization. However, as we open wallet-selector to new wallets, there is a fine balancing act of ensuring the user funds are safe and that the wallets provided by third party are compliant to wallet standards and security standards while we decentralize. The list that you see included here (sender is not part of the list) may change and consist our current understanding of wallet code bases what we have good understanding of wallet security. We will continuously evaluating this list as the space grows. 

I believe I shared on the PR you had raised. Here's a criteria for wallets https://github.com/near/wallet-selector/blob/main/CONTRIBUTING.md and we will also be listening closely to community on the feedback for the wallets listed. Thanks.

agileurbanite:
This PR negates the original intent as well, we want DApp developers not to have to re-deploy everytime a new list of recommended wallets is introduced.  This pushes the complexity of DApp developers to understand that a new list has been authored and will need to re-deploy their entire web app just to update wallet selector with a new list of wallets.

Ideally, DApp developers can:
* Implement wallet selector with a default list
* Override said default list through includes or excludes or custom array

hope this makes sense.

janewang:
>  This pushes the complexity of DApp developers to understand that a new list has been authored and will need to re-deploy their entire web app just to update wallet selector with a new list of wallets.
@agileurbanite What was the list that you are suggesting to put out to the DApp developer in the original context? Whatever the list it: it is either a) the entire list, b) some wallets of the packages, or c) no wallets. Since the original ticket asked for a list of wallets, I think it is either a) or b). I failed to see a conservative approach b) is a bad idea. I do not think the original ask of updating wallets during runtime has fully assessed the security risks.

MaximusHaximus:
@kujtimprenkuSQA Removing the magic strings and 'whitelist' argument is a good change, but we do still want the `recommended-wallets` package to expose a _single function_ like `setupRecommendedWallets` that will return the value that should be passed in as the `modules` argument to `setupWalletSelector()`.  This function is the 'sane defaults' that we want to encourage developers to use; without it, as @lostpebble pointed out, there is really no difference between the implementation here and just importing the individual packages themselves and manually passing them in as `modules`.  We still want the interface to look like the example code you had in #376:
```javascript
  network: "testnet",
  modules: await setupRecommendedWallets(),
});
```
If someone wanted default wallets plus one of their own, they could do something like:
```javascript
const myExampleWallet = {...};
  network: "testnet",
  modules: [...(await setupRecommendedWallets()), myExampleWallet],
});
```

MaximusHaximus (changes requested):
@kujtimprenkuSQA Removing the magic strings and 'whitelist' argument is a good change, but we do still want the `recommended-wallets` package to expose a _single function_ like `setupRecommendedWallets` that will return the value that should be passed in as the `modules` argument to `setupWalletSelector()`.  This function is the 'sane defaults' that we want to encourage developers to use; without it, as @lostpebble pointed out, there is really no difference between the implementation here and just importing the individual packages themselves and manually passing them in as `modules`.  We still want the interface to look like the example code you had in #376:
```javascript
  network: "testnet",
  modules: await setupRecommendedWallets(),
});
```
If someone wanted default wallets plus one of their own, they could do something like:
```javascript
const myExampleWallet = {...};
  network: "testnet",
  modules: [...(await setupRecommendedWallets()), myExampleWallet],
});
```

lostpebble:
> If someone wanted default wallets plus one of their own, they could do something like:
> 
> ```js
> const myExampleWallet = {...};
> const selector = await setupWalletSelector({
>   network: "testnet",
>   modules: [...(await setupRecommendedWallets()), myExampleWallet],
> });
> ```

This API feels a little convoluted when you want to add extra wallets and opens the door for weirdness (example, adding a wallet twice because it happens to be in the "recommended" package and a separate package- its not very obvious _what_ exactly is in the recommended wallets package).

I still don't really understand why this is necessary- "**Recommended Wallets**" could easily be a section in the readme with an example of how to import them. Also we're unsure why MyNearWallet is included by default- especially when they are now not an "official" wallet anymore and since they are a carbon copy of the previous Near Wallet, I expect they still have a known (and quite serious in my eyes) security issue (https://twitter.com/smsunarto/status/1555382740228247554). We've done research in this area in our own wallet, and it seems a purely "web wallet" will never be completely secure because of these storage issues.

As a final point, I feel like this whole extra package is going to discourage DApps from adding other "non-recommended" wallets. I feel like a lot of people would just implement this package and call it a day- without taking the time to do research on other wallets, since they're not "recommended" anyway.

packages/core/src/lib/services/wallet-modules/wallet-modules.service.ts
```diff
@@ -283,8 +283,9 @@ export class WalletModules {
         continue;
       }
 
+      // Skip duplicated module.
       if (modules.some((x) => x.id === module.id)) {
-        throw new Error("Duplicate module id detected: " + module.id);
+        continue;
```
MaximusHaximus, line 288: Curious as to why this was removed - should we not still be alerting someone if they configure the selector with multiple instances of the same module -- unique `id` values are important because they are used as keys on an object

packages/default-wallets/src/lib/default-wallets.ts
```diff
@@ -0,0 +1,22 @@
...
+import type { WalletModuleFactory } from "@near-wallet-selector/core";
+
+export const setupDefaultWallets = async (): Promise<
+  Array<WalletModuleFactory>
+> => {
+  return [
+    setupMyNearWallet(),
+    setupLedger(),
+    setupWalletConnect({
+      projectId: "c8cb6204543639c31aef44ea4837a554",
```
MaximusHaximus, line 13: hmm -- where did this projectId come from?

MaximusHaximus (changes requested):
Just two questions here

Also, I am curious about the inclusion of WalletConnect as a default; I was under the impression there are not yet any NEAR wallets supporting WC in the wild cc: @janewang

packages/core/src/lib/services/wallet-modules/wallet-modules.service.ts
kujtimprenkuSQA: If we add 2 wallets with the same id the last one this way will be skipped.
 If we don't add `continue` here the initialization of the wallet selector in `setupWalletSelector` will fail with the above error.
This was done to prevent failing of `setupWalletSelector` when we want to override one of the `default wallets`.

packages/default-wallets/src/lib/default-wallets.ts
kujtimprenkuSQA: This is coming from here: https://cloud.walletconnect.com/sign-in .
I have mentioned it a few times on Slack that if we add WalletConnect to the default wallets it will require some params.

kujtimprenkuSQA:
> Also, I am curious about the inclusion of WalletConnect as a default; I was under the impression there are not yet any NEAR wallets supporting WC in the wild

@MaximusHaximus , that's correct there's no official wallet that supports NEAR + WalletConnect.

packages/core/src/lib/services/wallet-modules/wallet-modules.service.ts
kujtimprenkuSQA: Developers may want to override one of the default wallets in the list to add custom options in the setup function of that specific wallet.

MaximusHaximus:
@kujtimprenkuSQA I had a chat with @janewang and she confirmed that we won't be shipping with WalletConnect in the initial default list.  I pushed a commit to remove it in the interest of efficiency/time.  I think this PR is ready to ship.

MaximusHaximus (approved):
:+1: Per my comment, I pushed a final commit to remove WalletConnect from the defaults for now.  This is ready to go otherwise I think!

kujtimprenkuSQA:
@MaximusHaximus, thank you very much for the help and the review in this PR.

AmmarHumackicSQA (approved):

PR #438, opened 2022-09-02
[meteor] Update SDK version

lostpebble:
# Description

Minor version update for Meteor SDK, fixes some dependencies which were left out of previous versions.

# Checklist:

# Type of change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "@jscutlery/semver": "^2.25.2",
     "@ledgerhq/hw-transport": "6.27.1",
     "@ledgerhq/hw-transport-webhid": "6.27.1",
-    "@meteorwallet/sdk": "0.4.0",
+    "@meteorwallet/sdk": "^0.6.0",
     "@nightlylabs/connect-near": "0.0.10",
     "@walletconnect/qrcode-modal": "2.0.0-alpha.20",
     "@walletconnect/sign-client": "^2.0.0-beta.102",


lostpebble:
Sorry for another one so soon @kujtimprenkuSQA - just realized that some dependencies were not included in our SDK package. I'm not sure why it even worked before without them, but I've included them now to be sure.

kujtimprenkuSQA:
@lostpebble I think you can run `yarn install` in your branch which will update the `yarn.lock`  I think it should be fine to do that in this PR and then commit those changes too, this way we will have the same `yarn.lock` as you have locally if that makes sense.

lostpebble:
Okay cool, I've added the `yarn.lock` into the PR. The reason I tried to remove it last time as well was because I was using a different version of Yarn I think- and the entire lock file was changed (was way too big to even show the diff on Github).

But it seems now that I'm using an older version again, it is only a small change this time.

kujtimprenkuSQA:
Yeah, it can happen sometimes btw this is the version of yarn on my machine `1.22.19`.
NX might read from the cache if there are no changes in the source code from the last build but you can skip cache by adding the `--skip-nx-cache` flag.

All packages
`yarn build:all --skip-nx-cache`

Single package:
`yarn nx run meteor-wallet:build --skip-nx-cache`

Will approve and merge this PR 👍

kujtimprenkuSQA (approved):

PR #460, opened 2022-09-15
Near Snap - MetaMask Snap to support NEAR network

BeroBurny:
# Description

NEAR Snap extends the capability of MetaMask to interact with the NEAR network.
* [Snaps](https://metamask.io/snaps/) is a system that allows anyone to expand the capabilities of MetaMask safely.
* [Snaps](https://metamask.io/snaps/) is pre-release software, To try Snaps, install [MetaMask Flask](https://metamask.io/flask).

# Checklist:


kujtimprenkuSQA:
Hey, @BeroBurny thank you for submitting this PR.

Can you make sure that `Near Snap` follows these criteria:
https://github.com/near/wallet-selector/blob/main/CONTRIBUTING.md#listing-criteria-for-third-party-wallet-on-wallet-selector

Could you please explain how to use this wallet: how do we create an account or import an existing one (i.e kujtimprenku.testnet)?

irubido:
> Hey, @BeroBurny thank you for submitting this PR.
> 
> Can you make sure that `Near Snap` follows these criteria: https://github.com/near/wallet-selector/blob/main/CONTRIBUTING.md#listing-criteria-for-third-party-wallet-on-wallet-selector
> 
> Could you please explain how to use this wallet: how do we create an account or import an existing one (i.e kujtimprenku.testnet)?

Hi @kujtimprenkuSQA,
Near Snap follows the criteria linked above.
It uses Metamask Flask extension [https://metamask.io/flask/](url) that support snaps, it derives Near address from metamask seed. 
Importing and exporting is the same as with Metamask, through extension.
I'll update readme to be more specific about Metamask interaction.

kujtimprenkuSQA:
Hey, @BeroBurny.

I will share my experience with `Near Snap` and I also have some questions.

For a user it is very unclear what is happening under the hood when using this wallet, I downloaded and installed the extension then created a new "Ethereum" account it was not clear how this account will be used just when I tried to sign in via Wallet Selector then I noticed a message **429380d6c79d10d1b7cd5b457153d8f433e9dec37e29bf52467cc1a30bdd74c0** does not exist this account was somehow created based on the "Ethereum" account created through the extension but because it had 0 NEAR on it the sign in did not work what's strange is that I never saw that account id anywhere in the Extension indicating that this is the account id based on the Ethereum account.

Questions/ comments:

- There's no way to know if the above account id is `testnet` or `mainnet` (near) account.
- The network in the extension shows as `Ethereum Mainnet` even though I was able to signIn and signOut with a near account (possibly testnet). - How do we add a NEAR testnet or mainnet network?
- When I try to sign and send a transaction the UI is not showing any useful information (like the amount, action type, text message, gas price etc).
- When you connect/signIn to a website through MetaMask on the left side there's a button that says `Connected` but in this case even though I sing in successfully it still says `Not Connected`
- I created a second Ethereum account and switching to that account does nothing when I sign in via Wallet Selector it still selects the above account id which was created for my first account.
- I transferred 5 NEAR from `kujtimprenku.testnet` to `429380d6c79d10d1b7cd5b457153d8f433e9dec37e29bf52467cc1a30bdd74c0` but they don't show anywhere in the `Near Snap` shows as **0 ETH $0.00 USD**
- How does signOut work? - I click the signout button and switch the account in the extension then try to sign in it still selects the first account I created there.
- Assets and Activity tabs in the extension don't show any information.

Did anything change in the source code of `Near Snap`?

Today I tried to test how this would work if I exported the account first and then removed the extension after that I re-installed it and imported the account but this wallet does not work anymore I see this error when I click Connect on the wallet's UI:
Errors-Flask

AmmarHumackicSQA:
Hey @BeroBurny @irubido I just wanted to check the status regarding the work on this. Do you guys have any updates or any questions? Hvala :wink:

danforbes:
Hi, @AmmarHumackicSQA - I am the ChainSafe project manager for our work on the NEAR snap. this Pull Request was made as part of a contract that was signed by ChainSafe and the NEAR Foundation. I agree with many of the concerns raised by @kujtimprenkuSQA, but in many/most cases they speak to limitations of the snap system itself. Snaps are still in beta so it's possible that as the snap ecosystem matures the concerns that have been brought up will be addressed, but at this time it's not within the scope of work (or in many cases even the scope of possibility) for ChainSafe developers to address the issues that have been brought up. Please let me know if you'd like to set up some time to chat about this and we'll figure out a way to make that work.

kujtimprenkuSQA:
Hey, @danforbes I have noticed two potential issues and I am unable to use this Wallet (snap)

- When `Metamask` is installed  then this error shows up, and the wallet selector fails to initialize (disabling Metamaks makes the error go away):

near-snap-error

- I am unable to sign in with this Wallet, this error shows up for me (this is the same error I mentioned on September 26 see the [comment](https://github.com/near/wallet-selector/pull/460#issuecomment-1257687689)):

near-snap-sign-in-error

I just installed **MetaMask Flask DEVELOPMENT** and created a new Account, and tried to sign in with Wallet Selector, is there any extra step/work I need to do to successfully signIn in to the guest-book example with Near Snap?

danforbes:
@kujtimprenkuSQA - this is a known bug that occurs when both MetaMask and MetaMask Flask are installed as extensions in the same browser. Based on the first screenshot you shared, that seems to be the cause of the problem you're experiencing.

gutsyphilip:
> d on the first screenshot yo



> @kujtimprenkuSQA - this is a known bug that occurs when both MetaMask and MetaMask Flask are installed as extensions in the same browser. Based on the first screenshot you shared, that seems to be the cause of the problem you're experiencing.

@danforbes Should we expect an update that'd fix these or is this out of the domain of Snaps?

danforbes:
@gutsyphilip the answer to that question is out of my domain as I am not associated with the MetaMask team; I am associated with the team at ChainSafe that implemented the Near snap, with the support of the Near Foundation.

packages/near-snap/src/lib/near-snap.ts
```diff
@@ -0,0 +1,110 @@
...
+    async signOut() {
+      // Nothing to do here!!
+    },
+
+    async getAccounts() {
+      return await getSnapAccounts(isDev, network);
+    },
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    async verifyOwner() {},
```
kujtimprenkuSQA, line 36: Can we throw an error if this method is not implemented in `Near Snap`?

```ts
async verifyOwner() {
    logger.log("NearSnap:verifyOwner", { message });
    
    throw new Error(`Method not supported by ${metadata.name}`);
}
```

packages/near-snap/README.md
```diff
@@ -0,0 +1,62 @@
...
+## Installation and Usage
+
+The easiest way to use this package is to install it from the NPM registry, this package requires `near-api-js` v0.44.2 or above:
+
+```bash
+# Using Yarn
+yarn add near-api-js@^0.44.2
+
+# Using NPM.
+npm install near-api-js@^0.44.2
```
kujtimprenkuSQA, line 15: Can you remove the `@^0.44.2` part from the installation section and just leave it like this:

```bash
# Using Yarn
yarn add near-api-js

# Using NPM.
npm install near-api-js
```

We have changed the range and this way it will mean that the latest version of near-api-js will be installed in a dApp.

packages/near-snap/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@near-wallet-selector/near-snap",
+  "version": "7.0.2",
+  "peerDependencies": {
+    "near-api-js": "^0.44.2"
+  }
```
kujtimprenkuSQA, line 6: Can you update the version to the latest version of the wallet selector, currently `v7.3.0`

Can you update the range of the peerDependency to this, because now we support `v1.x` of `near-api-js`:

```json
"peerDependencies": {
    "near-api-js": "^0.44.2 || ^1.0.0"
 }
```

packages/near-snap/src/lib/near-snap.ts
```diff
@@ -0,0 +1,110 @@
...
+  return {
+    async signIn() {
+      // Sign in to My Wallet for access to account(s).
+      await enable(isDev);
+
+      return await getSnapAccounts(isDev, network);
+    },
+
+    async signOut() {
+      // Nothing to do here!!
```
kujtimprenkuSQA, line 28: Why is this function empty?
How do we sign out from an account in Near Snap, and later on signIn into another account?

packages/near-snap/src/lib/near-snap.ts
```diff
@@ -0,0 +1,110 @@
...
+    },
+  };
+};
+
+export interface NearSnapParams {
+  iconUrl?: string;
+}
+
+export const setupNearSnap = ({
+  iconUrl = "./assets/near-snap-icon.png",
```
kujtimprenkuSQA, line 89: Can you convert the icon/image to base64 as we do in other packages and resize it since it's more than 200kb?

packages/near-snap/src/lib/near-snap.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    return {
+      id: "near-snap",
+      type: "injected",
+      metadata: {
+        name: "NearSnap",
+        description: null,
+        iconUrl,
+        deprecated: false,
```
kujtimprenkuSQA, line 103: Can you make the `deprecated` option configurable in the setupNearSnap, check other wallets how it's done?

packages/near-snap/src/lib/types.ts
```diff
@@ -0,0 +1,52 @@
+import { Action } from "@near-wallet-selector/core";
+
+declare global {
+  interface Window {
+    ethereum: {
+      isMetaMask: boolean;
+      request: <T>(
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        request: SnapRpcMethodRequest | { method: string; params?: Array<any> }
```
kujtimprenkuSQA, line 9: Can you please try adding the correct type here instead of the `any` type?

examples/react/contexts/WalletSelectorContext.tsx
```diff
@@ -56,10 +57,11 @@ export const WalletSelectorContextProvider: React.FC<{
         setupMathWallet(),
         setupNightly(),
         setupMeteorWallet(),
-        setupWelldoneWallet(),
+        setupNearSnap(),
```
kujtimprenkuSQA, line 60: Can you please add the `setupNearSnap()` function in the `Angular` example too?

kujtimprenkuSQA (commented):
Hey, @irubido thank you for updating the PR.
Left a few comments and questions, also looks like there are some linting issues/errors to check them try running:
` yarn nx run near-snap:lint`

packages/nearfi/src/lib/nearfi.ts
```diff
@@ -166,8 +166,8 @@ const NearFi: WalletBehaviourFactory<InjectedWallet> = async ({
     },
 
     async verifyOwner({ message }) {
-      logger.log("NearFi:verifyOwner", { message });
-
+      logger.log("NearSnap:verifyOwner", { message });
```
kujtimprenkuSQA, line 169: This PR is about `Near Snap` why is the change made in the `NearFi` wallet?

Please revert this change.

packages/near-snap/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@near-wallet-selector/near-snap",
+  "version": "v7.3.0",
+  "peerDependencies": {
+    "near-api-js": "^0.44.2 || ^1.0.0"
+  }
+}
```
kujtimprenkuSQA, line 7: We have recently made some updates in the `package.json` of each package to add some additional meta data.
Can you please align this with the other packages?

packages/near-snap/src/lib/near-snap.ts
```diff
@@ -0,0 +1,113 @@
...
+    return {
+      id: "near-snap",
+      type: "injected",
+      metadata: {
+        name: "NearSnap",
+        description: null,
+        iconUrl,
+        deprecated,
+        downloadUrl: "https://metamask.io/flask/",
+        available: await isMetaMaskAvailable(),
```
kujtimprenkuSQA, line 108: The `await isMetaMaskAvailable()` is blocking the execution.

Could you please fix this?

packages/near-snap/src/lib/types.ts
```diff
@@ -0,0 +1,58 @@
+import type { Action } from "@near-wallet-selector/core";
+
+declare global {
+  interface Window {
+    ethereum: {
+      isMetaMask: boolean;
+      request: <T>(
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
kujtimprenkuSQA, line 8: Is this comment needed?

packages/near-snap/src/lib/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export type GetSnapsResponse = {
+  [k: string]: {
+    permissionName?: string;
+    id?: string;
+    version?: string;
+    initialPermissions?: { [k: string]: unknown };
+  };
+};
```
kujtimprenkuSQA, line 18: The same `GetSnapsResponse` interface exists in 2 files, can you remove this and use only the one from `types.ts`?

packages/near-snap/src/lib/near-snap.ts
```diff
@@ -0,0 +1,113 @@
...
+const NearSnapWallet: WalletBehaviourFactory<InjectedWallet> = async ({
+  options,
+  provider,
+  store,
+}) => {
+  const network = options.network.networkId as "testnet" | "mainnet";
+
+  return {
+    async signIn() {
+      // Sign in to My Wallet for access to account(s).
```
kujtimprenkuSQA, line 22: Is this comment needed?

kujtimprenkuSQA (commented):
Currently, **Near Snap** is blocking the execution of the Wallet Selector when `setupNearSnap()` is added to the array of `modules` in the `setupWalletSelector()` for this reason, we can not manually test the functionality.

To reproduce: remove `node_modules` and `dist` install, build, and run everything again.

TODO:
- Address the comments.
- Merge with the latest changes of the `dev` branch.
- Make sure build, tests, and linting are working.
- Request a review again.

packages/near-snap/src/lib/near-snap.ts
kujtimprenkuSQA: Can we have a bit of info/insight on how `signOut` works here?

packages/near-snap/src/lib/near-snap.ts
kujtimprenkuSQA: Can you please take a look at this again?

packages/near-snap/src/lib/near-snap.ts
irubido: Hi,
I can't reproduce the issue of blocked execution, can you share more on what's causing the issue.

Do you have metamask flask installed, or regular metamask? You should have metamask flask for snap to work.

`isMetaMaskAvailable()` checks if metamask is installed, and calls metamask flask exclusive method 
(
```
await window.ethereum.request({
    method: "wallet_getSnaps",
  });
```
) 
to check if installed metamask is flask version.
Either way if it fails it just returns false if there is no metamask flask installed, and should not block rest of the app.

kujtimprenkuSQA: Hey, I know Metamask and Metamask Flask should not be active at the same time.

I had this "Near Snap" added to the Metamask Flask before back then when I did the review for the PR but yesterday I could not even try this option because it's blocking the execution so here's what I did to start "fresh".

- Remove Metamask
- Remove Metamast Flask
- Install again Metamask Flask

Here's what's happening:

blocked

After some more investigation somewhere down the line the `await isMetaMaskAvailable()` is calling a function `await getWalletSnaps()` which is just hanging there and never resolving.

I thought this could be an issue just for me but other team members tried it and they had the same experience.

irubido: For me loader spins for 5 seconds and then shows Log in. Snap works as intended afterwards.

What are the console errors?
Is the metamask flask version `10.23.2-flask.0`?

kujtimprenkuSQA: Hey, @irubido I have this version of Metamask Flask `10.23.2-flask.0` but I just found out what the issue is.

Somehow the having [MetaMask Flask](https://chrome.google.com/webstore/detail/metamask-flask-developmen/ljfoeinjpaedjfecbmggjgodbgkmjkjk) and [Coin98](https://chrome.google.com/webstore/detail/coin98-wallet/aeachknmefphepccionboohckonoeemg) extensions installed at the same time is blocking the execution.

I disabled the `Coin98` just for now and tested the `Near Snap` now everything is working I could:
`signIn`, `signOut`, `signAndSendTransaction(s)` with and without deposit.

Coin98 has been added to the wallet selector a few months ago.

Could you please investigate what's the conflict between `Coin98` and `Near Snap`?

irubido: Interesting, I'll take a look into it

BeroBurny: coin98 overrides `winsow.etherium` so the request is going to coin98 wallet, not to meta mask, and coin98 does not throw an error on not existing RPC call, but just hang out forever

BeroBurny: coin98 overrides `winsow.ethereum` so the request is going to coin98 wallet, not to meta mask, and coin98 does not throw an error on not existing RPC call, but just hangs out forever

BeroBurny: did some research with Coin98 wallet and they actually change methods and leave object unchanged
image
the issue is that there is a variable `isMetaMask` but the app is actually communicating with Coin98 wallet

Did some "hacking" to resolve this issue. (for curios one added promise race timeout 1 sec)
Think that now is testable.

cc @kujtimprenkuSQA

kujtimprenkuSQA: @BeroBurny thank you for taking the time to add the fix. Will test it out first thing in the morning.

kujtimprenkuSQA: Sorry for the late response, this looks much better at least the execution is not being blocked.

Even this way `Near Snap` is living somehow under the shadow of `Coin98`, if **Coin98** is installed this means the **Near Snap** will not work and will show the `WalletNotInstalled` view inaccurately.

We have informed the Coin98 team and we hope they prioritize this issue.

In the meantime can you resolve these 2 comments:
https://github.com/near/wallet-selector/pull/460#discussion_r1048523512
https://github.com/near/wallet-selector/pull/460#discussion_r1048536752

packages/near-snap/src/lib/near-snap.ts
BeroBurny: > Why is this function empty? 

There is no dedicated signout method that can be called programmatically, and as we do not store anything in the state there is no need to clean. Only `signIn` is important to request authorization for the page.

> How do we sign out from an account in Near Snap, and later on signIn into another account?

Currently, snaps do not have implemented a multi-account feature like a regular MetaMask but hope it will be possible to do in future

packages/near-snap/src/lib/near-snap.ts
BeroBurny: Step into contact with the MetaMask team to find a solution to that problem.
Meantime did some research additional to find an acceptable solution for a problem, but non without unreadable magic (like editing static `*.html`)

BeroBurny: Step into contact with the MetaMask team to find a solution to that problem.
Meantime did some research additional to find an acceptable solution for a problem, but non without unreadable magic (like editing static `*.html`)

packages/near-snap/src/lib/near-snap.ts
kujtimprenkuSQA: Thank you, for the explanation.

danforbes:
@kujtimprenkuSQA - are there any pending tasks related to this PR for the ChainSafe team?

kujtimprenkuSQA:
Hey, @danforbes the PR looks good to me,  there are no pending tasks from the Near Snap team.

kujtimprenkuSQA (dismissed):

irubido:
Updated breaking changes for metamask flask v1.25 released today

kujtimprenkuSQA:
> Updated breaking changes for metamask flask v1.25 released today

@irubido after this last update we're seeing an error when we try to signIn:

Screenshot 2023-02-20 155021

**Version
10.25.0-flask.0**

kujtimprenkuSQA:
Is there a "right" way to upgrade these snaps?

The way I did it was by removing the previous version and trying to connect with the new one, now it's working.

kujtimprenkuSQA (approved):

packages/near-snap/src/lib/near-snap.ts
kujtimprenkuSQA: @BeroBurny & @irubido thank you again for submitting this PR and resolving all the comments.

We have decided to merge this PR but wanted to let you know the response we got from the `Coin98` team.

Coin98 overrides `window.ethereum` to facilitate their user's connection to dApps. The current solution for this is to turn `off` the override, following either of these ways:

1. Through the "Override Wallet" feature on Coin98 Extension (Home > Settings > Override Wallet). Turning off the button means allowing the other wallet to override Coin98 and vice versa.
index

2. Through Chrome's "Manage Extensions" section > Deactivate Coin98 when wishing to use other wallets.

We will document this after merging this PR.

PR #464, opened 2022-09-19
Fix deploy on gh-pages

HarisSQA:
# Description

I had fix issue with angular deploy and added gh-pages into devDependecies. 

# Checklist:

- [ x] I have performed a self-review of my own code
- [x ] My changes generate no new warnings

# Type of change.

- [x ] FIX - a PR of this type patches a bug.

### DIFF
--- a/examples/angular/project.json
+++ b/examples/angular/project.json
@@ -138,7 +138,7 @@
     "deploy": {
       "executor": "@nrwl/workspace:run-script",
       "options": {
-        "script": "gh-pages -d dist/examples/angular"
+        "script": "gh-pages -d ./dist/examples/angular"
       }
     }
   },
--- a/package.json
+++ b/package.json
@@ -145,6 +145,7 @@
     "jest-mock-extended": "^2.0.6",
     "jest-preset-angular": "^12.2.0",
     "nx": "14.4.2",
+    "gh-pages": "^4.0.0",
     "prettier": "^2.7.1",
     "react-test-renderer": "18.2.0",
     "sass": "1.52.3",


examples/angular/project.json
```diff
@@ -138,7 +138,7 @@
     "deploy": {
       "executor": "@nrwl/workspace:run-script",
       "options": {
-        "script": "gh-pages -d dist/examples/angular"
+        "script": "gh-pages -d ./dist/examples/angular"
```
kujtimprenkuSQA, line 141: I don't think there's a need to change the path from an `absolute` to a `relative` here.

Do you have `gh-pages` installed globally? - if so the script will not work for other developers.

We need to install it locally: `yarn add -D gh-pages` and run the commands for building and deploying the angular example which are found in Contributing file.

HarisSQA: I think we need this, because it is execute from wallet-select/angular so without "./" path would be wallet-select/angular/dist/angular/. 

You can try without it but it will throw "src" error.

kujtimprenkuSQA: The command for deploying to gh-pages `yarn nx deploy angular` does not work for me.

not-found

kujtimprenkuSQA: Sorry was confused because `yarn.lock` did not change even though you've added `gh-pages` as a devDependency.

Run: `yarn install` this will update yarn.lock and push those changes.

We use `yarn` in this project instead of `npm` so we need to keep that in mind,

Not everyone has those extensions installed the commands in contributing.md should work fine.
I still think the path needs to be an absolute path and you can run these successfully from the root.


```bash
yarn nx build angular # build the angular example
yarn nx deploy angular # deploys angular example to gh-pages
```

PR #519, opened 2022-10-21
add neth to wallet-selector/packages

mattlockyer:
Adds NETH to wallet-selector core packages.

NETH is published on NPM as `neth` as well.

Documentation, source and the installation app can be found here: https://www.npmjs.com/package/neth


kujtimprenkuSQA:
Hey, @mattlockyer thank you for submitting this PR.

It might be still a work in progress but I just want to point out that dependencies in an NX project need to be added in the root `package.json`, this means you need to add/install dependencies in the root level so you don't need to go to the package /packages/neth and install them.

NX will detect any imports and will automatically add these dependencies during build time you might find them at /dist/packages/neth/package.json aftter you run the`yarn build:all` or a specific script command.

The base branch needs to be changed from `main` to `dev` it's explained here how the workflow goes:

PR #551, opened 2022-11-09
Typing Error for Content.tsx in the React Example

thalasith:
Added an interface to fix up the typing error in Content.tsx so that target.elements exists / is correct to the specific example.


kujtimprenkuSQA (commented):
Thank you for submitting this PR.

Can you please change the base branch from `main` to `dev` as it's shown in the docs:

examples/react/components/Content.tsx
```diff
@@ -181,13 +181,21 @@ const Content: React.FC = () => {
...
+  interface InputElements extends EventTarget {
+    elements: {
+      fieldset: HTMLFieldSetElement;
+      message: HTMLInputElement;
+      donation: HTMLInputElement;
+      multiple: HTMLInputElement;
+      button: HTMLButtonElement;
+    };
+  }
+
```
kujtimprenkuSQA, line 193: While this fixes the issue for the current elements of the form, if in the future we add a new element to the form then we would need to update this interface too.

I would suggest that we use a more "generic" type here, declare it at the top of the `Context.tsx` file below other imports:

```ts
export type Submitted = SubmitEvent & {
  target: { elements: { [key: string]: HTMLInputElement } };
};
```

Then we will need to make these changes (line 195):

```ts
async (e: Submitted ) => {}
```

And then align it in the `onSubmit` prop of the form

```jsx
<Form
  account={account}
  onSubmit={(e) => handleSubmit(e as unknown as Submitted)}
/>
 ```

thalasith: I don't know if the above generic solution will work for this case as fieldset is set as an "HTMLFieldSetElement".

Is there a way to add this into the elements? Would this also assume that all future elements will strictly be HTMLInputElements?

kujtimprenkuSQA: Well, the reason `HTMLInputElement ` would work fine, here is that in the code we access only these properties: `disabled`, `value`,  `checked`, `focus()` looking at the docs here these properties will be present: https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement so typescript would not complain.

Since the goal of this PR is to fix the types for the current form elements and I like the idea of setting the correct types I am going to approve it.

kujtimprenkuSQA:
Could you please fix the linting issues?

kujtimprenkuSQA:
Hey, @thalasith could you please fix the lint issues so the  GH actions run without any errors?

thalasith:
hopefully that fixes it - first time helping in OSS! Thank you for your patience with me.

kujtimprenkuSQA:
Can you please discard your changes in `yarn.lock` and remove the `package.lock.sjon` file?

kujtimprenkuSQA:
@thalasith Thank you for taking the time to open this PR.

Not to take so much of your time, decided to address this fix in a new PR: https://github.com/near/wallet-selector/pull/558 I have mentioned this PR there so thank you for your contribution.

PR #554, opened 2022-11-10
feat: Added wallet order optimization

amirsaran3:
# Description
- Added `optimizeWalletOrder` option when setting up wallet-selector. 
- By default it is `true`.
- It will prioritize last signed in wallets, then installed wallets over not installed and deprecated wallets.

# Checklist:

# Type of change.

### DIFF
--- a/packages/core/src/lib/constants.ts
+++ b/packages/core/src/lib/constants.ts
@@ -1,6 +1,5 @@
 export const PACKAGE_NAME = "near-wallet-selector";
-export const LAST_SIGNED_IN_WALLET =
-  "near-wallet-selector:last_signed_in_wallet";
+export const RECENTLY_SIGNED_IN_WALLETS = "recentlySignedInWallets";
 
 export const CONTRACT = "contract";
 export const PENDING_CONTRACT = "contract:pending";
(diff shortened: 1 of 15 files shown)


packages/core/src/lib/constants.ts
```diff
@@ -1,4 +1,6 @@
 export const PACKAGE_NAME = "near-wallet-selector";
+export const LAST_SIGNED_IN_WALLET =
+  "near-wallet-selector:last_signed_in_wallet";
```
kujtimprenkuSQA, line 3: I think we can remove the package name and the column `near-wallet-selector:` here otherwise it will be duplicated when you create a new `jsonStorage` :
```ts
  const jsonStorage = new JsonStorage(this.storage, PACKAGE_NAME);
```
It will look something like this in localstorage: 
`near-wallet-selector:near-wallet-selector:last_signed_in_wallet`

Would you also consider setting it as `lastSignedInWallet` instead of `last_signed_in_wallet` it will align with the way you set it in the store (state)?

```ts
export const LAST_SIGNED_IN_WALLET = "lastSignedInWallet";
```

kujtimprenkuSQA (approved):

PR #520, opened 2022-10-21
add neth

mattlockyer:
@kujtimprenkuSQA changes have been made.

NETH repo for development, docs and NETH install page:
https://github.com/NearDeFi/neth/

[neth.app](https://neth.app/) live NETH installation page.

PR #556, opened 2022-11-14
Fix/neth naj1.1.0

mattlockyer:
This addresses the issue raised in https://github.com/near/wallet-selector/pull/520#issuecomment-1311795095


packages/neth/src/lib/neth-lib.ts
```diff
@@ -104,8 +104,9 @@ export const initConnection = ({
   keyStore = new BrowserLocalStorageKeyStore();
   near = new Near({
     ...network,
-    deps: { keyStore },
+    keyStore,
```
kujtimprenkuSQA, line 107: Can you please keep just this change and revert other changes?

I am sorry if my comment on the other PR was confusing, at the moment we have been just testing wallet-selector with **NAJ v1.1.0** but we did not decide if we should update the version yet.

I checked and this change works fine with `v0.44.2` too.

PR #567, opened 2022-11-19
[WIP] feat: core package i18n zh translation

think-in-universe:


think-in-universe:
I think the Chinese translation are ready now.

AmmarHumackicSQA:
> I think the Chinese translation are ready now.
> 
> Please notice usually we'll support both Simplified Chinese and Traditional Chinese to make it usable for different districts (including Hong Kong, Taiwan). But I think it would be good to start with the Simplified Chinese first.

Thank you for your effort @think-in-universe !

HarisSQA (approved):
Everything seems good.

HarisSQA:
Hey @think-in-universe , we noticed that this part is missing 
`"connectingMessage": {
        "injected": "Confirm the connection in the extension window",
        "browser": "Confirm the connection in the wallet after redirect",
        "hardware": "Confirm the connection in the ledger device",
        "bridge": "Confirm the connection in the wallet"
} `
 and if you can help with translating it as we are not sure if it would be correct if we just use the google translate. Thanks!

PR #575, opened 2022-11-22
feat: add new methods to WalletBehaviour types

andy-haynes:
# Description

This PR extends the `InjectedWalletBehaviour` and `BrowserWalletBehaviour` types to include new, optional, mutually exclusive methods:

- `buildImportAccountsUrl()` (browser wallets) returns the URL for the import accounts page on the wallet. The encrypted JSON string will be appended onto it as a fragment.
- `importAccountsInSecureContext({ accounts: Array<{ accountId: string, publicKey: string }> })` takes an array of accounts to import, represented by objects with an `accountId` and `privateKey`. It returns a `Promise<void>` though there could be some value in utilizing return codes for injected wallets. For now it's not worth taking on the added complexity and instead we can rely on thrown exceptions to indicate failure.

I've opted to have these methods take a set of accounts rather than a single account because:
- the URL flow already supports multiple accounts in a single link
- the implementer should determine how best to import one or more accounts

# Checklist:

# Type of change.

### DIFF
--- a/packages/core/src/lib/wallet/wallet.types.ts
+++ b/packages/core/src/lib/wallet/wallet.types.ts
@@ -100,10 +100,6 @@ export interface BrowserWalletSignInParams extends SignInParams {
   failureUrl?: string;
 }
 
-export interface AccountImportEncryptedParams {
-  encryptedAccountJson: string;
-}
-
 export interface BrowserWalletSignAndSendTransactionParams
   extends SignAndSendTransactionParams {
   callbackUrl?: string;
@@ -117,9 +113,7 @@ export interface BrowserWalletSignAndSendTransactionsParams
 export type BrowserWalletBehaviour = Modify<
   BaseWalletBehaviour,
   {
-    importAccountsByEncryptedUrl?(
-      params: AccountImportEncryptedParams
-    ): Promise<void>;
+    buildImportAccountsUrl?(): string;
     importAccountsInSecureContext?: never;
     signIn(params: BrowserWalletSignInParams): Promise<Array<Account>>;
     signAndSendTransaction(
@@ -155,7 +149,7 @@ export interface AccountImportSecureContextParams {
 export type InjectedWalletBehaviour = Modify<
   BaseWalletBehaviour,
   {
-    importAccountsByEncryptedUrl?: never;
+    buildImportAccountsUrl?: never;
     importAccountsInSecureContext?(
       params: AccountImportSecureContextParams
     ): Promise<void>;
--- a/packages/core/src/lib/wallet/wallet.types.ts
+++ b/packages/core/src/lib/wallet/wallet.types.ts
@@ -117,10 +117,10 @@ export interface BrowserWalletSignAndSendTransactionsParams
 export type BrowserWalletBehaviour = Modify<
   BaseWalletBehaviour,
   {
-    importAccountByEncryptedUrl?(
+    importAccountsByEncryptedUrl?(
       params: AccountImportEncryptedParams
     ): Promise<void>;
-    importAccountInSecureContext?: never;
+    importAccountsInSecureContext?: never;
     signIn(params: BrowserWalletSignInParams): Promise<Array<Account>>;
     signAndSendTransaction(
       params: BrowserWalletSignAndSendTransactionParams
@@ -155,8 +155,8 @@ export interface AccountImportSecureContextParams {
 export type InjectedWalletBehaviour = Modify<
   BaseWalletBehaviour,
   {
-    importAccountByEncryptedUrl?: never;
-    importAccountInSecureContext?(
+    importAccountsByEncryptedUrl?: never;
+    importAccountsInSecureContext?(
       params: AccountImportSecureContextParams
     ): Promise<void>;
   }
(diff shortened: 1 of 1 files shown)


packages/core/src/lib/wallet/wallet.types.ts
```diff
@@ -113,6 +117,10 @@ export interface BrowserWalletSignAndSendTransactionsParams
 export type BrowserWalletBehaviour = Modify<
   BaseWalletBehaviour,
   {
+    importAccountsByEncryptedUrl?(
+      params: AccountImportEncryptedParams
+    ): Promise<void>;
```
hcho112, line 122: Just wondering, is `importAccountsByEncryptedUrl` need to be a promise? 
What I assumed was that this function should receive an argument(`AccountImportEncryptedParams`) and return string (url).

hcho112, line 122: one question here, so what is expected returned promise like?
eg 1. url link. (string url, https://somewallet.io/batch-import#${hash})
eg 2. function call `window.open(url, '_blank')`

hcho112: As outcome of our discussion, once you made the change to replace `Promise<void>` to `string`, I will approve them :)

packages/core/src/lib/wallet/wallet.types.ts
```diff
@@ -113,6 +113,8 @@ export interface BrowserWalletSignAndSendTransactionsParams
 export type BrowserWalletBehaviour = Modify<
   BaseWalletBehaviour,
   {
+    buildImportAccountsUrl?(): string;
```
hcho112, line 116: arguments to this interface has been removed from this commit. Is there a reason why?

packages/core/src/lib/wallet/wallet.types.ts
```diff
@@ -135,7 +137,24 @@ export type InjectedWalletMetadata = BaseWalletMetadata & {
...
+}
+
+export interface AccountImportSecureContextParams {
+  accounts: Array<AccountImportData>;
+}
+
+export type InjectedWalletBehaviour = Modify<
+  BaseWalletBehaviour,
+  {
+    buildImportAccountsUrl?: never;
```
hcho112, line 152: Can we rethink about applying `buildImportAccountsUrl` as never on `InjectedWallet` ?
Currently Meteor wallet is defined as injected wallet on `wallet-selector` but their approach towards to wallet migration behaves just like browser wallet. (So based on our flow, Meteor wallet should be using `buildImportAccountsUrl`.)

Furthermore, some wallet may support more than one type of interface, (eg. practically, wallet like Meteor wallet, does have both browser AND extension wallet where they can support both `buildImportAccountsUrl` and `importAccountsInSecureContext` interfaces)

andy-haynes: > Currently Meteor wallet is defined as injected wallet on wallet-selector but their approach towards to wallet migration behaves just like browser wallet.

Hmm do we know why they're configured as an injected wallet if they function as a browser wallet? I don't think it's worth making exceptions for misconfigured wallets but if there's something we've missed here then it's important to incorporate that into our model.

It is extremely risky to have an injected wallet send a link, as injected wallets use unencrypted data but anything in a URL must be encrypted. We need to make this as straightforward and secure as possible, so intermixing wallet types and making one-off exceptions breaks our proposed implementation.

> Furthermore, some wallet may support more than one type of interface, (eg. practically, wallet like Meteor wallet, does have both browser AND extension wallet where they can support both buildImportAccountsUrl and importAccountsInSecureContext interfaces)

This is a limitation of `wallet-selector` AFAIK. Since each wallet only has exactly one type, each package represents a wallet + mode (of which most wallets will only have one). I don't think this is a relationship we want to change though; if I'm selecting Meteor as my target wallet then isn't it important that I select whether I want the mobile wallet or browser wallet? I'd imagine they have quite different implementation considerations.

So at the moment I would prefer to have wallets implement a separate package for a different mode, assuming that there is not significant overlap between the same wallet in different modes. If this ends up being untenable for wallet builders then we have some potentially major architecture decisions to make.

packages/core/src/lib/wallet/wallet.types.ts
andy-haynes: Yes, it no longer needs parameters:

>  `buildImportAccountsUrl()` (browser wallets) returns the URL for the import accounts page on the wallet. The encrypted JSON string will be appended onto it as a fragment.

The idea is they return the URL for importing accounts in the current network, and we append the encrypted data after a `#`. Now the wallet implementation never needs encrypted data and we've enforced that encrypted data is added as a fragment.

I believe we discussed this last week but I don't think I updated the drafted proposal.

hcho112: I assumed that what you are saying is that calling `buildImportAccountsUrl()` will return `https://www.example-wallet.io&someparams#` and we append encrypted data next after `#`?

But would it enforce the url expression to always have `#` at the end of the url. Would it be more developer friendly if we just take encrypted string as argument of a function and return completed url?

```
(encryptedData) => `https://www.example-wallet.io#${encryptedData}&someparams`
```

hcho112 (dismissed):
LGTM, will address remaining issues on other PR

hcho112:
maybe shall we change the target from `main` to `dev` and merge?

andy-haynes:
> maybe shall we change the target from `main` to `dev` and merge?

Fixed, thanks!

hcho112 (approved):
LGTM

PR #578, opened 2022-11-24
feat: Initial implementation of importAccountsInSecureContext on Sender Wallet

hcho112:
# Description

This PR contains initial implementation for adding importAccountsInSecureContext interface for Sender Wallet. [WEP-212](https://pagodaplatform.atlassian.net/browse/WEP-212)

# Checklist:

# Type of change.

[WEP-212]: https://pagodaplatform.atlassian.net/browse/WEP-212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/lib/options.types.ts
+++ b/packages/core/src/lib/options.types.ts
@@ -1,7 +1,7 @@
 export type NetworkId = "mainnet" | "testnet";
 
 export interface Network {
-  networkId: NetworkId;
+  networkId: string;
   nodeUrl: string;
   helperUrl: string;
   explorerUrl: string;
--- a/packages/core/src/lib/options.types.ts
+++ b/packages/core/src/lib/options.types.ts
@@ -1,7 +1,7 @@
 export type NetworkId = "mainnet" | "testnet";
 
 export interface Network {
-  networkId: string;
+  networkId: NetworkId;
   nodeUrl: string;
   helperUrl: string;
   explorerUrl: string;
--- a/packages/sender/src/lib/sender.ts
+++ b/packages/sender/src/lib/sender.ts
@@ -279,10 +279,12 @@ const Sender: WalletBehaviourFactory<InjectedWallet> = async ({
         });
     },
 
-    importAccountsInSecureContext(accountData) {
+    async importAccountsInSecureContext({ accounts: accountData }) {
       if (window && window.near && window.near.isSender) {
+        // TODO: apply encryption here when it's implemented
+        const encryptedAccountData = JSON.stringify(accountData);
         window.near.batchImport({
-          keystore: accountData,
+          keystore: encryptedAccountData,
           network: options.network.networkId,
         });
       } else {
--- a/packages/sender/src/lib/sender.ts
+++ b/packages/sender/src/lib/sender.ts
@@ -278,6 +278,20 @@ const Sender: WalletBehaviourFactory<InjectedWallet> = async ({
           return res.response;
         });
     },
+
+    importAccountsInSecureContext(accountData) {
+      if (window && window.near && window.near.isSender) {
+        window.near.batchImport({
+          keystore: accountData,
+          network: options.network.networkId,
+        });
+      } else {
+        window.open(
+          "https://chrome.google.com/webstore/detail/sender-wallet/epapihdplajcdnnkdeiahlgigofloibg?utm_source=chrome-ntp-icon",
+          "_blank"
+        );
+      }
+    },
   };
 };
 
--- a/packages/sender/src/lib/sender.ts
+++ b/packages/sender/src/lib/sender.ts
@@ -282,14 +282,9 @@ const Sender: WalletBehaviourFactory<InjectedWallet> = async ({
     async importAccountsInSecureContext({ accounts }) {
       if (window.near && window.near.isSender) {
         window.near.batchImport({
-          keystore: JSON.stringify(accounts),
+          keystore: accounts,
           network: options.network.networkId,
         });
-      } else {
-        window.open(
-          "https://chrome.google.com/webstore/detail/sender-wallet/epapihdplajcdnnkdeiahlgigofloibg?utm_source=chrome-ntp-icon",
-          "_blank"
-        );
       }
     },
   };
--- a/packages/sender/src/lib/sender.ts
+++ b/packages/sender/src/lib/sender.ts
@@ -279,12 +279,10 @@ const Sender: WalletBehaviourFactory<InjectedWallet> = async ({
         });
     },
 
-    async importAccountsInSecureContext({ accounts: accountData }) {
-      if (window && window.near && window.near.isSender) {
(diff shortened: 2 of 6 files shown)


packages/core/src/lib/options.types.ts
```diff
@@ -1,7 +1,7 @@
 export type NetworkId = "mainnet" | "testnet";
 
 export interface Network {
-  networkId: string;
+  networkId: NetworkId;
```
kujtimprenkuSQA, line 4: Why are we changing the type here from `string` to `NetworkId`?

This will limit the network id for custom networks  to just either "testnet" or "mainnet"

```ts
    network: {
      networkId: "localnet",
      nodeUrl: "http://127.0.0.1:52993",
      helperUrl: "http://127.0.0.1:52997",
      explorerUrl: "http://127.0.0.1:53009",
      indexerUrl: "http://127.0.0.1:52997",
    },
    modules: [setupNearWallet()],
  });
```

Typescript would throw an error: **_TS2322: Type '"localnet"' is not assignable to type 'NetworkId'_**

hcho112: Initially my thoughts were, leaving `networkId` as `string` isn't really helpful to anyone. And then saw `NetworkId` type above so I assume that this was left as backlog. (My mistake)

If we are officially supporting `localnet` and any other network, should be update `NetworkId` type and use it instead of string? (eg. `export type NetworkId = "mainnet" | "testnet" | "localnet";`)

kujtimprenkuSQA: The idea is that we have presents only for testnet and mainnet  so NetworkId is used for that reason, if a developer wishes to use another network they can provide an object of type Network.

hcho112: Thanks for the correction, reverted to where it was

andy-haynes (changes requested):
Looking good! Mostly just need to remove references to encrypted data

packages/sender/src/lib/sender.ts
```diff
@@ -278,6 +278,20 @@ const Sender: WalletBehaviourFactory<InjectedWallet> = async ({
           return res.response;
         });
     },
+
+    async importAccountsInSecureContext({ accounts }) {
+      if (window.near && window.near.isSender) {
+        window.near.batchImport({
+          keystore: JSON.stringify(accounts),
```
andy-haynes, line 285: The PR calling this method expects data as an [array of arrays](https://github.com/near/near-wallet/pull/2957/files#diff-594a571a3658085ce173558b02d6f1a52304c8116640d608beb737e60883a5c3R90) so we can just map it to the legacy format here for now.

If we haven't already we'll need to reach out to Sender about the data no longer being encrypted; we can start the conversation on that linked PR. This implementation won't work until that change is made at the very least, but they may also want to change the `keystore` parameter name and potentially match our interface of an array of objects for each account ID/private key pair.

packages/sender/tsconfig.json
```diff
@@ -4,7 +4,8 @@
     "forceConsistentCasingInFileNames": true,
     "strict": true,
     "noImplicitReturns": true,
-    "noFallthroughCasesInSwitch": true
+    "noFallthroughCasesInSwitch": true,
+    "resolveJsonModule": true,
```
andy-haynes, line 8: Is this needed for this PR? Just curious, I didn't notice any JSON file imports

andy-haynes (changes requested):
Great work, just some comments around trying to match the current implementation proposal by Sender and getting opinionated about how we handle uninstalled extensions. I think we can move ahead with this without waiting for them to remove the encryption requirement since this functionality should start working once removed.

packages/sender/src/lib/sender.ts
```diff
@@ -278,6 +278,20 @@ const Sender: WalletBehaviourFactory<InjectedWallet> = async ({
...
+          network: options.network.networkId,
+        });
+      } else {
+        window.open(
```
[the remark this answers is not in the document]
kujtimprenkuSQA: Just to inform you each module has an [available](https://github.com/near/wallet-selector/blob/main/packages/sender/src/lib/sender.ts#L312) property in the `metadata` object in case of Injected wallets we call the `isInstalled()` function which checks if an extension is installed during setup and sets the value for the `available` property.

Then in UI, you can check it this way: https://github.com/near/wallet-selector/blob/main/packages/modal-ui/src/lib/components/Modal.tsx#L124-L130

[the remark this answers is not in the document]
hcho112: I think the modal ui already handles uninstalled wallets nicely on it's own way. 
image
I'm implementing on top of modal-ui and also have layer of logic to check for interface implemented. If not, will show this:
image

I'm happy to remove `else` part of the logic.

packages/sender/tsconfig.json
hcho112: This is needed to be done in order to get the sender test running. Without this, upon running the sender test, it complains about undefined path on `en.json`

packages/sender/src/lib/sender.ts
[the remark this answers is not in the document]
andy-haynes: Perfect, thank you @kujtimprenkuSQA and @hcho112 🎉

packages/sender/tsconfig.json
andy-haynes: Good to know, thanks!

packages/sender/src/lib/sender.ts
andy-haynes: @hcho112 I was thinking we could pass in the format that is currently expected by `batchImport`, but let's get clarification from Sender Wallet either way before we decide on an implementation. That way we don't end up doing it twice 😄

packages/sender/src/lib/sender.ts
```diff
@@ -278,6 +278,15 @@ const Sender: WalletBehaviourFactory<InjectedWallet> = async ({
           return res.response;
         });
     },
+
+    async importAccountsInSecureContext({ accounts }) {
+      if (window.near && window.near.isSender) {
+        window.near.batchImport({
```
hcho112, line 284: Based on [comments here](https://github.com/near/wallet-selector/pull/650#issuecomment-1396948993) this part need to be re-written.

kujtimprenkuSQA: Hi, @hcho112 I think we can leave this part for now, in my opinion, what was mentioned in this comment https://github.com/near/wallet-selector/pull/650#issuecomment-1396948993 is out of the scope of this PR.

kujtimprenkuSQA:
@hcho112  we will be closing this PR in favor of the new one created by Sender's team: https://github.com/near/wallet-selector/pull/859

PR #594, opened 2022-12-06
HERE Wallet | Async API

AZbang:
HERE Wallet now supports asynchronous interaction like Meteor.

We are using our internal js-sdk. You can read more details here:
https://github.com/here-wallet/js-sdk

You can also touch a demo here:
https://codesandbox.io/s/here-wallet-instant-app-6msgmn


HarisSQA:
- Please change the base branch from **main** to **dev**.
- Looks like the SSL certificate has expired on your wallet could you fix it.
- **signAndSendTransaction** is now asking for an approval on the wallet even for transactions that don't have any deposit in them, is this intended?

HarisSQA:
@AZbang , It's seems that we still have issue with SSL. Do you have any infos when this could be fixed?

HarisSQA (approved):

PR #604, opened 2022-12-15
Add `signMessage` method to Wallet Selector

kujtimprenkuSQA:
# Description

- Added and exported interfaces for `signMessage` in `core` package.
- Added `signMessage` method in `BaseWalletBehaviour`
- Added `signMessage` method to all wallets (currently it only throws until wallets add the implementation).

The work on this PR is based on the NEP: https://github.com/near/NEPs/pull/413/

Remaining work:

- [ ] Update examples to call and test `signMessage`.
- [ ] Verify the signMessage response in examples (closes: https://github.com/near/wallet-selector/issues/434).
- [ ] Update WalletConnect examples to include the `signMessage` function too.
- [ ] Once the NEP is merged reach out to Wallet builders to add support for `signMessage` in their wallets.

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings

# Type of change.

- [ ] FIX - a PR of this type patches a bug.
- [ ] FEATURE - a PR of this type introduces a new feature.
- [ ] BUILD - a PR of this type introduces build changes.
- [ ] CI - a PR of this type introduces CI changes.
- [ ] DOCS - a PR of this type introduces DOCS improvement.
- [ ] STYLE - a PR of this type introduces style changes.
- [ ] REFACTOR - a PR of this type introduces refactoring.
- [ ] PERFORMANCE - a PR of this type introduces performance changes.
- [ ] TEST - a PR of this type adds more tests.
- [ ] CHORE - a PR introduces other changes than the specified above.

PR #617, opened 2022-12-24
here-wallet: add signMessage

AZbang:
Implement signMessage Standart for HERE Wallet
Standart draf: https://github.com/near/NEPs/pull/413

`signMessage` sign hash of `NEP0413:` + `{"message":"hi","nonce":[0,...,31],"receiver":"myapp.com"}` string by full access key and return signature, then verifies the signature and verifies that the public key is indeed the full access key on the blockchain.

If verification fails, the method returns an error.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "@angular/platform-browser": "~14.0.0",
     "@angular/platform-browser-dynamic": "~14.0.0",
     "@angular/router": "~14.0.0",
-    "@here-wallet/core": "^1.1.0",
+    "@here-wallet/core": "^1.2.1",
     "@jscutlery/semver": "^2.27.2",
     "@ledgerhq/hw-transport": "6.27.1",
     "@ledgerhq/hw-transport-webhid": "6.27.1",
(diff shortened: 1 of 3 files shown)


amirsaran3 (approved):

kujtimprenkuSQA:
Hey, @AZbang thank you for adding support for the `signMessage` on HERE Wallet.

We have created this PR https://github.com/near/wallet-selector/pull/604 to officially add support for `signMessage` function to the API of all wallets in Wallet Selector.

After merging with the current changes in `dev` branch I noticed a few things that need to be changed in order HERE to fully align with the NEP.

- The params need to be signed in an alphabetical order, I can see only this in your [SDK](https://github.com/here-wallet/js-sdk/blob/main/src/wallet.ts#L264) but I assume you have signed this the same way in the wallet too.
- The return type of the params for `signature` and the `publicKey` need to be aligned with the NEP: https://github.com/gutsyphilip/NEPs/blob/nep/wallet-werifyOwner/neps/nep-0413.md#output-interface

You can have a look at the example mentioned in the NEP: https://github.com/gutsyphilip/NEPs/blob/nep/wallet-werifyOwner/neps/nep-0413.md#references

PR #631, opened 2023-01-08
Add Arabic translation

SSekaiking:
# Description

Added Arabic language


DamirSQA:
Hi @SSekaiking  new section "Export Account" is added recently, so it would be great if there's translation [for it](https://github.com/near/wallet-selector/blob/main/packages/core/src/lib/locale/en.json#L58) as well. Would you be able provide it? Thanks.

PR #632, opened 2023-01-09
feat: add support of xdefi wallet

hinterlist:
# Description

Add support of [XDEFI Wallet](https://www.xdefi.io/) which can be installed from [Chrome Store](https://chrome.google.com/webstore/detail/xdefi-wallet/hmeobnfnfcmdkdcmlblgagmfpfboieaf?hl=en)

# Checklist:


tsconfig.base.json
```diff
@@ -77,6 +77,10 @@
       "@near-wallet-selector/coin98-wallet": [
         "packages/coin98-wallet/src/index.ts"
       ],
+      "@near-wallet-selector/xdefi": [
+        "packages/xdefi/src/index.ts"
+      ]
```
amirsaran3, line 82: Build is failing because of this extra parenthesis. Can you please remove it.

amirsaran3 (changes requested):
Can you run `yarn lint` and fix the linting issues.

tsconfig.base.json
hinterlist: good point

amirsaran3:
Thanks for the quick update @hinterlist,

After creating an account with XDEFI wallet when I go to the react guest-book example I get this error:
image

Also can you add XDEFI wallet to the angular example.

amirsaran3:
Can you also add the XDEFI wallet to the angular example.

amirsaran3:
When I tested on another machine I can login, but I can't sign transactions.

I get this error: 
image

hinterlist:
Let me update dependencies and fix the issues that you've dicovered. Integration was made couple month ago and pushed just recently, so I suspect something has changed since then.

AmmarHumackicSQA:
> Let me update dependencies and fix the issues that you've dicovered. Integration was made couple month ago and pushed just recently, so I suspect something has changed since then.

Hey @hinterlist when can we expect that? We plan to do a release tomorrow, and we could include this PR if the mentioned can be addressed by then. Thanks!

hinterlist:
> When I tested on another machine I can login, but I can't sign transactions.
> 
> I get this error: image

Done debuging, the reason of this error is that we do not support testnet in prod build, though itenral rpc used for example is assuming testnet. Trying to figure out what's the best way to perform proper testing, few options I see here, and I don't won't to break your normal flow, so first option is to add support of testnet for prod, the second option would be do test on mainnet. Another option, I can send build that will use testnet instead of mainnet.

hinterlist:
Support of testnet is ready, I'll ping you when app store will be updated with a new build of extension.

hinterlist:
A new version of the wallet in the store, it support testnet.

amirsaran3:
@hinterlist Now when I try to sign in I get this. 
image
No console errors appear.

hinterlist:
@amirsaran3 have you switch to testnet as I've described above?

amirsaran3 (approved):

PR #634, opened 2023-01-10
feat: WEP-217 Introducing account-export package

hcho112:
# Description

This PR contains introduction of new export-selector-ui package to wallet selector.

The main purpose of export-selector-ui is to provide a way to easily and safely migrate given account(s) to listed third party wallets that also supports import/export interfaces. 

It is expected to have few more PRs needed prior to release this to production. 

Here are few screenshots of the work:

# Checklist:

# Type of change.


packages/export-selector-ui/src/lib/components/ExportAccount.tsx
```diff
@@ -0,0 +1,351 @@
...
+  accountId: string;
+  publicKey: string;
+}
+
+const getAccountType = async ({
+  provider,
+  accountId,
+  publicKey,
+}: getAccountTypeProps) => {
+  const result = await provider
```
andy-haynes, line 89: 💄 I think this should either return the promise chain or use a `try`/`catch` with `await`, mixing is a little awkward when the result of the chain is all that needs to be returned, e.g.

```js
return provider.query(...);
```

or

```js
try {
  const { permission } = await provider.query(...);
  return permissionToType(permission);
} catch {
  return ...
}
```

packages/export-selector-ui/src/lib/components/ExportAccount.tsx
```diff
@@ -0,0 +1,351 @@
...
+interface getAccountBalanceProps {
+  provider: providers.Provider;
+  accountId: string;
+}
+
+const getAccountBalance = async ({
+  provider,
+  accountId,
+}: getAccountBalanceProps) => {
+  const result = await provider
```
andy-haynes, line 119: 💄 as above

packages/export-selector-ui/src/lib/components/ExportAccount.tsx
```diff
@@ -0,0 +1,351 @@
...
+  >([]);
+  const [disabledAccounts, setDisabledAccounts] = useState<
+    Array<ExportAccountData>
+  >([]);
+
+  //@ts-ignore
+  const [exportInterfaces, setExportInterfaces] = useState<{
+    buildImportAccountsUrl?: BrowserWalletBehaviour["buildImportAccountsUrl"];
+    importAccountsInSecureContext?: InjectedWalletBehaviour["importAccountsInSecureContext"];
+    // TODO:  once WEP-219 is completed, implement following interface
```
andy-haynes, line 165: What would you think about leaving out QR code references until we're ready to implement the interface?

packages/export-selector-ui/src/lib/components/ExportAccount.tsx
```diff
@@ -0,0 +1,351 @@
...
+    buildQrCode?: () => void;
+  }>({});
+  useEffect(() => {
+    const getExportInterfaces = async () => {
```
andy-haynes, line 169: Is this just calling synchronous methods? If so could this be inlined in the `useEffect`?

andy-haynes (changes requested):
Awesome work Danny! Thank you for including the example as part of this PR 💖 

I have some relatively minor feedback and may have more after I try out the Meteor integration on top of this

packages/export-selector-ui/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "@near-wallet-selector/export-selector-ui",
```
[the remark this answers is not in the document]
hcho112: hmm right, `account-export` sounds much better. 
I will make updates on name change. (Require bit of changes)

packages/export-selector-ui/src/lib/components/ExportAccount.tsx
hcho112: you mean remove them from this PR?
I have implemented them with assumption that we will implement QR code interfaces soon. But if we can't give certainty at this stage, I think it is wise to delete them to avoid confusion. 

If we are removing QR code references, do you prefer to remove them completely or just in this component?

packages/export-selector-ui/src/lib/components/ExportAccount.tsx
hcho112: there is `await wallet.wallet()` below

packages/export-selector-ui/package.json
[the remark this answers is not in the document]
hcho112: I have made the change, but I will only push after your reviews get all resolved. (Only 2 reviews left to confirm)

examples/react/components/ExportContent.tsx
```diff
@@ -0,0 +1,20 @@
...
+const ExportContent: NextPage = () => {
+  const { ExportModal } = useExportAccountSelector();
+  return (
+    <Fragment>
+      <button onClick={() => ExportModal.show()}>Open Modal</button>
+      <p>
+        Export modal assist users to migrate given accounts to one of listed
+        third party wallets. Depends on the type of wallet selected, users may
+        required to copy the given passhrase and store it safely until the
+        migration is completed.
```
andy-haynes, line 14: I think something like this might read more clear:

> The Export Accounts modal assists users in migrating their accounts to any Wallet Selector wallet supporting account imports. Any sensitive data in transit during this process will be encrypted with a randomly-generated password which the user will be required to copy down and use in the account import process on the target wallet.

scripts/release-packages.bash
```diff
@@ -23,3 +23,4 @@ npm publish dist/packages/nearfi --tag "${TAG}" --otp "${OTP}"
 npm publish dist/packages/opto-wallet --tag "${TAG}" --otp "${OTP}"
 npm publish dist/packages/welldone-wallet --tag "${TAG}" --otp "${OTP}"
 npm publish dist/packages/neth --tag "${TAG}" --otp "${OTP}"
+npm publish dist/packages/account-export --tag "${TAG}" --otp "${OTP}"
```
hcho112, line 26: I would like to know if this will publish the `account-export` package properly.

examples/react/contexts/WalletSelectorExportContext.tsx
```diff
@@ -0,0 +1,158 @@
...
+          },
+        }),
+      ],
+    });
+    /**
+     * Insert list of accounts to be imported here
+     * accounts: [{ accountId: "test.testnet", privateKey: "ed25519:..."}, ...]
+     */
+    const _modal = setupExportSelectorModal(_selector, {
+      accounts: [],
```
kujtimprenkuSQA, line 88: How will a user in a dApp set their accounts here?

It looks like only the dApp developer has control over this at the moment.

kujtimprenkuSQA (commented):
- When clicking an option that does not support account export the CLOSE button is missing.
- Is there any wallet that supports account export just to test out the functionality?

scripts/release-packages.bash
kujtimprenkuSQA: At the moment the release process is quite manual which means one of the team members will have to do it, once we bump the version and merge everything to `main` then locally we would do a clean build of all the packages and run this script.

For first-time scoped npm packages, we would need to add an extra flag `--access=public` before we publish it.

```bash
npm publish dist/packages/account-export --tag "${TAG}" --otp "${OTP}" --access=public
```

I will ping you on slack and explain everything.

packages/account-export/src/lib/index.tsx
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+import { createRoot } from "react-dom/client";
+import type { WalletSelector } from "@near-wallet-selector/core";
+
+import type { WalletSelectorModal, ExportSelectorOptions } from "./index.types";
+
+import { ExportSelector } from "./components/ExportSelector";
+
+const MODAL_ELEMENT_ID = "near-wallet-selector-modal";
```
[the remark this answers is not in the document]
hcho112: there is a reason why I was using the same id:
1. All CSS stylings are reflected based on element id. [ref](https://github.com/near/wallet-selector/blob/main/packages/modal-ui/src/lib/components/styles.css#L3) one of the main aim was to inherit/reuse existing css styling from modal-ui. (To provide consistent look and feel)
2. Chances of both modal being executed is not realistic. Normally, user will use modal-ui mostly once? upon authenticate to dApp. Likewise, they will only use account-export modal once upon exporting

If I change the id to something else, I will have to clone modal-ui's css into account-export and duplicate variable declaration part. (Let me know if you have a good suggestion on this)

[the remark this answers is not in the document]
kujtimprenkuSQA: Alright, let's not change it then the second point makes sense 👍 .

examples/react/contexts/WalletSelectorExportContext.tsx
hcho112: dApp dev can use account-export part of code along with their own implementation of ability to dynamically pass `accountId` and `privateKey`. 

In terms of plan moving forward, `account-export` will be implemented on `near-wallet`. Part of the code from `near-wallet` will be dynamically create list of accounts (and their private keys stored in local storage) and feed them into account-export code. 

I would expect dApp devs to use account-export same way as above.

hcho112:
> * Is there any wallet that supports account export just to test out the functionality?

you can test them on My Near Wallet. 
(need to manually pass `accounts`)

Once its successful, it will redirect to MNW as below:
image

FYI, actual import sequence on other wallet won't work as we are still in the middle of making encoding/decoding standardized.

packages/account-export/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@near-wallet-selector/account-export",
+  "version": "7.4.0",
+  "peerDependencies": {
+    "near-api-js": "^1.1.0"
+  }
```
kujtimprenkuSQA, line 6: I think we should set the range of NAJ the same as in other packages:
```json
  "peerDependencies": {
    "near-api-js": "^0.44.2 || ^1.0.0"
  }
```
The reason we still need `^0.44.2` is that this is the version that was used by Wallet Selector for the majority of the time and there could be dApps that haven't moved to the latest version of NAJ yet.

kujtimprenkuSQA (approved):

andy-haynes (approved):
🎉

PR #647, opened 2023-01-17
feat: Implementing encryption and decryption on account-export WEP-213

hcho112:
# Description
(Review this PR only after [prerequisite PR](https://github.com/near/wallet-selector/pull/634) gets merged)
This PR introduces following helper interfaces from `account-export` package.

`encryptAccountData` - Encrypt account data for browser wallet
`decryptAccountData` - Decrypt account data for browser wallet

Currently `My Near Wallet` contains implementation for how to use  `encryptAccountData` and `decryptAccountData` data. 

It uses [generate-password](https://www.npmjs.com/package/generate-password) package to generate complex 32 character passphrase. 
It uses [crypto-js](https://www.npmjs.com/package/crypto-js) package to apply AES encryption/decryption on account data.

As outcome of this PR:
image

Once this PR gets merged, we will notify third party wallet teams to implement the account import part of their wallet side according to the wallet type defined in `wallet-selector`.

# Checklist:

# Type of change.


packages/account-export/src/lib/components/Passphrase.tsx
```diff
@@ -18,16 +19,26 @@ export const Passphrase: React.FC<PassphraseProps> = ({
...
-    numbers: true,
-    strict: true,
-    lowercase: true,
-    uppercase: true,
-    symbols: true,
-  });
+  const [secretKey, setSecretKey] = useState("");
+
+  useEffect(() => {
+    const key = generator.generate({
```
andy-haynes, line 27: Could this go in `helpers/importAccount` instead? It would be nice to consolidate all methods involved in account encryption into a single module.

packages/account-export/src/lib/helpers/importAccount.spec.ts
```diff
@@ -0,0 +1,36 @@
+import type { AccountImportData } from "@near-wallet-selector/core";
```
andy-haynes, line 1: Do you think there would be value in testing invalid/unexpected inputs here? E.g. invalid accounts JSON, empty `secretKey` values, etc.

packages/account-export/src/lib/helpers/importAccount.ts
```diff
@@ -0,0 +1,37 @@
...
+export const encodeAccountData = (
+  accountsData: Array<AccountImportData>
+): string => encodeURIComponent(JSON.stringify(accountsData));
+
+export const decodeAccountData = (
+  encodedAccountsData: string
+): Array<AccountImportData> =>
+  JSON.parse(decodeURIComponent(encodedAccountsData));
+
+interface encryptAccountDataProps {
```
andy-haynes, line 14: 💄 `EncryptAccountDataProps`

packages/account-export/src/lib/helpers/importAccount.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export const encryptAccountData = ({
+  accountData,
+  secretKey,
+}: encryptAccountDataProps): CryptoJS.lib.CipherParams => {
+  return CryptoJS.AES.encrypt(JSON.stringify(accountData), secretKey);
+};
+
+interface decryptAccountDataProps {
```
andy-haynes, line 26: 💄 `DecryptAccountDataProps`

packages/account-export/src/index.ts
```diff
@@ -1,4 +1,10 @@
-export { setupExportSelectorModal } from "./lib";
+export {
+  setupExportSelectorModal,
+  encodeAccountData,
+  decodeAccountData,
+  encryptAccountData,
```
andy-haynes, line 5: Is there a good case for exporting the `encryptAccountData` method if it's being used internally by the modal workflow? My only concern is that developers might think they need to use that method when it's actually something we're already handling.

andy-haynes (changes requested):
Looks great! Just a few observations, mostly hoping we no longer need the `encode*` methods 🤞

packages/account-export/src/index.ts
hcho112: Totally agree, I will remove them from export

packages/account-export/src/lib/helpers/importAccount.ts
```diff
@@ -0,0 +1,37 @@
+import CryptoJS from "crypto-js";
+
+import type { AccountImportData } from "@near-wallet-selector/core";
+
+export const encodeAccountData = (
```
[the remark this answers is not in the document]
hcho112: yes we do. This is prerequisite step to convert account data to string + web friendly format. 

Do you see a reason why we should remove this?

packages/account-export/src/lib/helpers/importAccount.spec.ts
hcho112: validity of accounts are actually get filtered during account selection step. 
(Invalid accounts will become un-selectable) 

Will update encryption and decryption logic with exception handling

packages/account-export/src/lib/helpers/importAccount.ts
```diff
@@ -0,0 +1,62 @@
...
+    throw new Error("Secret key is required");
+  }
+  try {
+    return CryptoJS.AES.encrypt(JSON.stringify(accountData), secretKey);
+  } catch {
+    throw new Error("Unable to encrypt account data");
+  }
+};
+
+interface DecryptAccountDataProps {
```
kujtimprenkuSQA, line 34: Can we move the interfaces to the top of the file just below other imports?

It's been a practice in this project to keep functions and types/interfaces somehow separate.

kujtimprenkuSQA (commented):
This looks good to me, just left two minor comments.

packages/account-export/src/lib/helpers/importAccount.ts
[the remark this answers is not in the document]
andy-haynes: It was my understanding that the encrypted data would be base64-encoded and not require any URL-escaping.

> Do you see a reason why we should remove this?

I mentioned it because it is only invoked in a test. It's not clear to me why we need it if we are passing encrypted data in the URL.

packages/account-export/src/lib/helpers/importAccount.ts
```diff
@@ -0,0 +1,62 @@
...
+
+interface EncryptAccountDataProps {
+  accountData: Array<AccountImportData>;
+  secretKey: string;
+}
+
+export const encryptAccountData = ({
+  accountData,
+  secretKey,
+}: EncryptAccountDataProps): CryptoJS.lib.CipherParams => {
```
andy-haynes, line 23: Ideally this would just return the encrypted string and the decryption method would take an encrypted string as well. Since we're exposing parts of this in the API we should keep the interface as simple as possible.

Is there a compelling reason to have it return a `CryptoJS` object?

packages/account-export/src/lib/helpers/importAccount.ts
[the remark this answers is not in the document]
hcho112: Thanks for the correction. I misunderstood the use case of encode/decode account data. Will remove them on next commit

hcho112:
Thanks for the kind reviews @andy-haynes and @kujtimprenkuSQA  all raised issues has been addressed.

andy-haynes (dismissed):
Nice work!

kujtimprenkuSQA (dismissed):

hcho112:
Recently I have received security feedback on this PR.

One change I have made is to pull out `crypto-js` and replacing the encryption with `tweetnacl` package. (To avoid introducing a new package and make the wallet-selector heavy. Also in terms of encryption and decryption, `secretbox` from `tweetnacl` was good enough)

Another feedback I have received is that we should encourage receiving wallet to execute `history.replaceState()` after receiving the hash data to make sure history is not stored.

kujtimprenkuSQA (dismissed):

packages/account-export/src/lib/helpers/importAccount.ts
```diff
@@ -0,0 +1,86 @@
...
+    const base64DecryptedMessage = encodeUTF8(decrypted);
+    return JSON.parse(base64DecryptedMessage);
+  } catch {
+    throw new Error("Unable to decrypt account data");
+  }
+};
+
+export const generateSecretKey = (): string =>
+  generator.generate({
+    length: 32,
```
andy-haynes, line 80: Only 32 char-length strings work as values here, pretty sure this is why... Let's set this using the library's value https://github.com/dchest/tweetnacl-js/blob/master/README.md#naclsecretboxkeylength--32

```suggestion
    length: nacl.secretbox.keyLength,
```

andy-haynes (changes requested):
Couple minor things otherwise LGTM

andy-haynes (approved):

andy-haynes (approved):

PR #682, opened 2023-02-02
fix: Update generateSecretKey from account-export package

### LINKED ISSUE #680: Exporting accounts using secure URL format fails with TypeError (crypto module)
I've started testing out the account export implementation with Meteor today and ran into this issue:

```
Uncaught TypeError: crypto.randomBytes is not a function
    at getNextRandomValue (generate.js:11:24)
    at randomNumber (generate.js:23:13)
    at generate (generate.js:49:20)
    at module.exports.generate (generate.js:138:17)
    at generateSecretKey (index.js:4946:43)

Uncaught TypeError: Cannot read properties of undefined (reading 'length')
    at getNextRandomValue (generate.js:9:62)
    at randomNumber (generate.js:23:13)
    at generate (generate.js:49:20)
    at module.exports.generate (generate.js:138:17)
    at generateSecretKey (index.js:4946:43)
    at index.js:5005:17

The above error occurred in the <Passphrase> component:

    at Passphrase (http://localhost:4200/node_modules/.vite/deps/@near-wallet-selector_account-export.js?v=d2207208:4506:3)
    at ExportAccount (http://localhost:4200/node_modules/.vite/deps/@near-wallet-selector_account-export.js?v=d2207208:4713:3)
```

It appears to be an issue with the `generate-password` library (https://github.com/brendanashworth/generate-password) which is used to generate the secret key to encrypt the user's accounts and private keys for exports:

```ts
export const generateSecretKey = (): string =>
  generator.generate({
    length: nacl.secretbox.keyLength,
    numbers: true,
    strict: true,
    lowercase: true,
    uppercase: true,
    symbols: true,
  });
```

On their docs, they say that it is not browser compatible (since they `require()` a Node.js module), which means it can't serve our purposes here. They do mention a browser-compatible version of their library though: https://www.npmjs.com/package/generate-password-browser.

But- I'm not really sure why we are adding this as a dependency though? The Near API JS library already has the ability to generate random keys on the fly, right? :

```ts
import { KeyPair } from "near-api-js";

const randomKey = KeyPair.fromRandom("ed25519").getPublicKey().toString();
```

Could that not serve our purposes here?

hcho112:
# Description

Fixes #680 

This PR contains implementation for changing `generateSecretKey` interface of account-export.
1. Removed  `generate-password` package
2. Update `generateSecretKey` to use `tweetnacl.randomBytes`
3. Minor update on `importAccount.spec` tests

# Checklist:

# Type of change.

### DIFF
--- a/packages/account-export/src/lib/helpers/importAccount.ts
+++ b/packages/account-export/src/lib/helpers/importAccount.ts
@@ -1,4 +1,3 @@
-import generator from "generate-password";
 import nacl from "tweetnacl";
 import {
   decodeUTF8,
@@ -74,12 +73,7 @@ export const decryptAccountData = ({
   }
 };
 
-export const generateSecretKey = (): string =>
-  generator.generate({
-    length: nacl.secretbox.keyLength,
-    numbers: true,
-    strict: true,
-    lowercase: true,
-    uppercase: true,
-    symbols: true,
-  });
+export const generateSecretKey = (): string => {
+  const random = nacl.randomBytes(nacl.secretbox.keyLength);
+  return encodeBase64(random).substring(0, 32);
+};
--- a/packages/account-export/src/lib/helpers/importAccount.ts
+++ b/packages/account-export/src/lib/helpers/importAccount.ts
@@ -74,6 +74,6 @@ export const decryptAccountData = ({
 };
 
 export const generateSecretKey = (): string => {
-  const random = nacl.randomBytes(nacl.secretbox.keyLength);
-  return encodeBase64(random).substring(0, 32);
+  const random = nacl.randomBytes(24);
+  return encodeBase64(random);
 };
--- a/package.json
+++ b/package.json
@@ -99,7 +99,6 @@
     "core-js": "^3.6.5",
     "crypto-browserify": "^3.12.0",
     "ethers": "^5.7.2",
-    "generate-password": "^1.7.0",
     "is-mobile": "^3.1.1",
     "near-api-js": "^1.1.0",
     "near-seed-phrase": "^0.2.0",
--- a/packages/account-export/src/lib/helpers/importAccount.spec.ts
+++ b/packages/account-export/src/lib/helpers/importAccount.spec.ts
@@ -1,6 +1,10 @@
 import type { AccountImportData } from "@near-wallet-selector/core";
 
-import { encryptAccountData, decryptAccountData } from "./importAccount";
+import {
+  encryptAccountData,
+  decryptAccountData,
+  generateSecretKey,
+} from "./importAccount";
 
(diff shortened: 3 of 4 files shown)


packages/account-export/src/lib/helpers/importAccount.ts
```diff
@@ -74,12 +73,7 @@ export const decryptAccountData = ({
...
-    length: nacl.secretbox.keyLength,
-    numbers: true,
-    strict: true,
-    lowercase: true,
-    uppercase: true,
-    symbols: true,
-  });
+export const generateSecretKey = (): string => {
+  const random = nacl.randomBytes(nacl.secretbox.keyLength);
+  return encodeBase64(random).substring(0, 32);
```
hcho112, line 78: @andy-haynes I would like to get your feedback on here!
While I spent time on using `nacl` to generate 32 character secret key, I couldn't figure out and  ended up using `substring` to enforce the character size to 32. Do you know more elegant solution on this?

andy-haynes: The byte array contains wider (2^8) values than base 64 (2^6), so encoding in base 64 requires more symbols to represent the same amount of information. The length of the encoded string is always 44, with the last character being `=` for padding.  So the byte array of length 32 represents up to `(2^8)^32 = 256^32 = 1.2e77` values and a base 64 string of length 43 is `(2^6)^43 = 64^43 = 4.6e77`, making 43 the lowest number of base 64 characters you can use to represent all values in the byte array.

To get a 32 character base 64 string, you need a byte array of length `log_256(64^32) = 24` (https://www.wolframalpha.com/input?i=log+base+256+%2864%5E32%29). We should be able to rely on that to always produce strings of length 32 😄

andy-haynes: Also the byte array of length 32 always had a `=` padding character because 32 is not a multiple of 3. A byte array of length 24 is a multiple of 3 and so will never contain padding, so we will always make the most of our 32 characters.

https://en.wikipedia.org/wiki/Base64#Output_padding

> Because Base64 is a six-bit encoding, and because the decoded values are divided into 8-bit octets, every four characters of Base64-encoded text (4 sextets = 4 × 6 = 24 bits) represents three octets of unencoded text or data (3 octets = 3 × 8 = 24 bits). This means that when the length of the unencoded input is not a multiple of three, the encoded output must have padding added so that its length is a multiple of four. The padding character is =, which indicates that no further bits are needed to fully encode the input.

hcho112: @andy-haynes awesome! thank you for guidance and explanation!

andy-haynes (approved):
🚀

PR #683, opened 2023-02-03
chore: Bumped version to 7.6.2

hcho112:
# Description

Bumped version for all packages to v7.6.2

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings

# Type of change.

- [ ] FIX - a PR of this type patches a bug.
- [ ] FEATURE - a PR of this type introduces a new feature.
- [ ] BUILD - a PR of this type introduces build changes.
- [ ] CI - a PR of this type introduces CI changes.
- [ ] DOCS - a PR of this type introduces DOCS improvement.
- [ ] STYLE - a PR of this type introduces style changes.
- [ ] REFACTOR - a PR of this type introduces refactoring.
- [ ] PERFORMANCE - a PR of this type introduces performance changes.
- [ ] TEST - a PR of this type adds more tests.
- [x] CHORE - a PR introduces other changes than the specified above.

PR #685, opened 2023-02-03
Revert "chore: Bumped version to 7.6.2"

kujtimprenkuSQA:
Reverts near/wallet-selector#683


kujtimprenkuSQA:
The reason why we are reverting this PR is described here: https://github.com/near/wallet-selector/pull/683#issuecomment-1415506751

AmmarHumackicSQA (approved):

PR #688, opened 2023-02-03
fix: Connection issue with wallets (PublicKey)

HarisSQA:
# Description

# Checklist:

# Type of change.

### DIFF
--- a/packages/my-near-wallet/src/lib/my-near-wallet.ts
+++ b/packages/my-near-wallet/src/lib/my-near-wallet.ts
@@ -16,6 +16,7 @@ import type {
 } from "@near-wallet-selector/core";
 import { createAction } from "@near-wallet-selector/wallet-utils";
 import icon from "./icon";
+import edgeSSRLoader from "next/dist/build/webpack/loaders/next-edge-ssr-loader";
 
 export interface MyNearWalletParams {
   walletUrl?: string;
@@ -75,7 +76,6 @@ const MyNearWallet: WalletBehaviourFactory<
   { params: MyNearWalletExtraOptions }
 > = async ({ metadata, options, store, params, logger }) => {
   const _state = await setupWalletState(params, options.network);
-
   const getAccounts = async (): Promise<Array<Account>> => {
     const accountId = _state.wallet.getAccountId();
     const account = _state.wallet.account();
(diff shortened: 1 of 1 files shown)


packages/my-near-wallet/src/lib/my-near-wallet.ts
```diff
@@ -16,6 +16,7 @@ import type {
 } from "@near-wallet-selector/core";
 import { createAction } from "@near-wallet-selector/wallet-utils";
 import icon from "./icon";
+import edgeSSRLoader from "next/dist/build/webpack/loaders/next-edge-ssr-loader";
```
kujtimprenkuSQA, line 19: Why is this import statement needed here?

packages/my-near-wallet/src/lib/my-near-wallet.ts
```diff
@@ -75,26 +76,36 @@ const MyNearWallet: WalletBehaviourFactory<
...
-            account.accountId,
-            options.network.networkId
-          )
-        ).toString(),
-      },
-    ];
+    const publicKeyTemp = await account.connection.signer.getPublicKey(
+      account.accountId,
+      options.network.networkId
+    );
```
kujtimprenkuSQA, line 89: It looks like `signIn` without a `contractId` is still not working for **MyNearWallet** the error is not showing but you can not sign in either:

I would suggest that we change the `getAccounts` function into something like this:

```ts
const getAccounts = async (): Promise<Array<Account>> => {
    const accountId = _state.wallet.getAccountId();
    const account = _state.wallet.account();
    if (!accountId || !account) {
      return [];
    }
    try {
      const publicKey = await account.connection.signer.getPublicKey(
        account.accountId,
        options.network.networkId      );
      return [
        {
          accountId,
          publicKey: publicKey.toString(),
        },
      ];
    } catch (e) {
      return [
        {
          accountId,
          publicKey: undefined,
        },
      ];
    }
  };
```

In the past, the `getAccounts` method was not `async` so the best we can do for now is check and catch any error while trying to get the `publicKey` when there are no errors we return the values for `accountId` and `publicKey` if there are errors for `publicKey` we return `undefined`.

kujtimprenkuSQA (changes requested):
Please check also other wallets since we added the `publicKey` in the Account interface the `getAccounts` has changed from a synchronous method to an asynchronous method.

For example Sender throws an error in this case if the wallet is `Locked`

packages/my-near-wallet/src/lib/my-near-wallet.ts
HarisSQA: I didn't realize that. Thank you!

packages/my-near-wallet/src/lib/my-near-wallet.ts
HarisSQA: Thank you for info and your example of code 👍

kujtimprenkuSQA (approved):

PR #689, opened 2023-02-06
feat: add account-export for WELLDONE

Yoon-Suji:
# Description

Add support for `buildImportAccountsUrl` in WELLDONE Wallet.
We also implemented the `importAccountsInSecureContext` method, but we decided to use the `buildImportAccountsUrl` for now because there is a little problem when calling the method without creating a wallet. If we solve the problem before the wallet-migration event starts, we plan to use the `importsAccountsInSecureContext` method.

# Checklist:

# Type of change.

### DIFF
--- a/packages/welldone-wallet/src/lib/welldone.ts
+++ b/packages/welldone-wallet/src/lib/welldone.ts
@@ -330,8 +330,27 @@ const WelldoneWallet: WalletBehaviourFactory<InjectedWallet> = async ({
       return results;
     },
 
-    buildImportAccountsUrl() {
-      return `https://docs.welldonestudio.io/contribution/202211-batch-import`;
+    async importAccountsInSecureContext({ accounts }) {
+      if (!_state.wallet) {
+        throw new Error("Wallet is not installed");
+      }
+      const privateKeys: Array<string> = [];
+      // use batch import
+      accounts.forEach(({ privateKey }) => {
+        if (privateKey.slice(0, 8) === "ed25519:") {
+          privateKeys.push(privateKey.slice(8));
+        } else {
+          privateKeys.push(privateKey);
+        }
+      });
+      const params = {
+        privateKey: privateKeys,
+        network: options.network.networkId,
+      };
+      await _state.wallet.request("near", {
+        method: "experimental:near:importPrivatekey",
+        params: [params],
+      });
     },
   };
 };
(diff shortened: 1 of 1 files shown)


packages/welldone-wallet/src/lib/welldone.ts
```diff
@@ -329,6 +329,29 @@ const WelldoneWallet: WalletBehaviourFactory<InjectedWallet> = async ({
 
       return results;
     },
+
+    buildImportAccountsUrl() {
+      return `https://docs.welldonestudio.io/contribution/202211-batch-import`;
+    },
+
+    async importAccountsInSecureContext({ accounts }) {
```
hcho112, line 337: Let's remove `importAccountsInSecureContext` on this PR if it is not ready from WELLDONE injected wallet side.

packages/welldone-wallet/src/lib/welldone.ts
```diff
@@ -355,6 +378,7 @@ export function setupWelldoneWallet({
           "https://chrome.google.com/webstore/detail/welldone-wallet/bmkakpenjmcpfhhjadflneinmhboecjf",
         deprecated,
         available: installed,
+        useUrlAccountImport: true,
```
hcho112, line 381: Let's remove `useUrlAccountImport` on this PR if it is not ready

hcho112 (changes requested):
Thank you for supporting the account export @Yoon-Suji 
Rest looks all good, but I would like to recommend to support either `buildImportAccountsUrl` or `importAccountsInSecureContext` but not both. 

For the sake of maintaining the codebase clean, we should aim to merge and push code if it is production ready. Refer to your description, currently `importAccountsInSecureContext` is not going to work right now, so ideally this part of implementation should be kept locally or in a seperate branch. 

When `importAccountsInSecureContext` is ready, I would recommend to create a PR that includes `importAccountsInSecureContext` and removes `buildImportAccountsUrl`.

Yoon-Suji:
Hi @hcho112 !
As you mentioned, I removed `importAccountsInSecureContext`, so WELLDONE Wallet only supports `buildImportAccountsUrl` method for now. I'll create a new PR when it is ready from the wallet side :)

packages/welldone-wallet/src/lib/welldone.ts
hcho112: my bad, we actually need this to be true in order to use `buildImportAccountsUrl` for injected wallet 😅

hcho112 (dismissed):
LGTM, I also ran the code locally and was able to export accounts successfully to WELLDONE wallet on testnet environment 👍

Yoon-Suji:
Hello @hcho112 ,
`importAccountsInSecureContext` is ready from the wallet and it's already been updated in Chrome Webstore! So I changed the method. Could you do the test again? Thank you :)

hcho112 (approved):
LGTM, `importAccountsInSecureContext` was working on local dev environment.

PR #726, opened 2023-03-10
Test coverage

amirsaran3:
`sender.spec.ts` doesn't work. I am not getting any errors so I am not sure how to fix it. I need some help with it.


kujtimprenkuSQA:
Add `"resolveJsonModule": true` to the [compilerOotions](https://github.com/near/wallet-selector/blob/dev/packages/sender/tsconfig.json#L3) of the `tsconfig.json` of the `sender` package.

Not sure why the `.json` files are being picked by NX when we run the tests, this should make `sender.spec.ts` work.

kujtimprenkuSQA (approved):

PR #728, opened 2023-03-14
Added custom language string for developers

HarisSQA:
# Description

# Checklist:

# Type of change.

### DIFF
--- a/packages/core/src/lib/translate/translate.ts
+++ b/packages/core/src/lib/translate/translate.ts
@@ -7,6 +7,8 @@ import vi from "../locale/vi.json";
 import hi from "../locale/hi.json";
 import ar from "../locale/ar.json";
 
+const devLanguage = "es"; // leave null or change into language code "en", "es"...
+
 const getLanguage = (languageCode: string) => {
   switch (languageCode) {
     case "en":
@@ -54,6 +56,9 @@ const findObjectPropByStringPath = (obj: any, prop: string): unknown => {
 export const translate = (path: string) => {
   let lang = window.navigator.languages ? window.navigator.languages[0] : null;
   lang = lang || window.navigator.language;
+  if (devLanguage !== null) {
+    lang = devLanguage;
+  }
 
   const languageCode = shortenLanguageCode(lang);
 


packages/core/src/lib/translate/translate.ts
```diff
@@ -7,6 +7,8 @@ import vi from "../locale/vi.json";
 import hi from "../locale/hi.json";
 import ar from "../locale/ar.json";
 
+const devLanguage = "es"; // leave null or change into language code "en", "es"...
```
DamirSQA, line 10: Maybe switch wariable name to "allowOnlyLanguage"

packages/core/src/lib/translate/translate.ts
DamirSQA, line 1: When finished, instruction on how to use should be added to the [languages  guide](https://github.com/near/wallet-selector/blob/main/packages/core/docs/guides/multilanguage-support.md)

DamirSQA (changes requested):

packages/core/src/lib/translate/translate.ts
```diff
@@ -54,6 +56,9 @@ const findObjectPropByStringPath = (obj: any, prop: string): unknown => {
 export const translate = (path: string) => {
   let lang = window.navigator.languages ? window.navigator.languages[0] : null;
   lang = lang || window.navigator.language;
+  if (devLanguage !== null) {
```
DamirSQA, line 59: Based on sugestion ` useAllowOnlyLanguage && (lang = allowOnlyLanguage);` can be used instead of `if{...} block

packages/core/src/lib/translate/translate.ts
HarisSQA: Sounds good!

packages/core/src/lib/translate/translate.ts
HarisSQA: My opinion is that `if{...}` block already does its function. For this method, you will need to add a new variable `useAllowOnlyLanguage=true`  and `allowOnlyLanguage`. So instead of change language only in one place `allowOnlyLanguage="es"`, developers would need to add also `useAllowOnlyLanguage=true`

DamirSQA (dismissed):

packages/core/src/lib/translate/translate.ts
```diff
@@ -7,6 +7,8 @@ import vi from "../locale/vi.json";
 import hi from "../locale/hi.json";
 import ar from "../locale/ar.json";
 
+const allowOnlyLanguage = null; // leave null or change into language code "en", "es"...
```
kujtimprenkuSQA, line 10: How are developers going to change the value of this `allowOnlyLanguage` variable on their dApp?

DamirSQA: @kujtimprenkuSQA, thanks for your comment. 
@HarisSQA   `allowOnlyLanguage`  should be put in function that is exported, so language as an argument could be passed to `translate.js`. from dApp.

HarisSQA: Done. Is now okay?

DamirSQA (changes requested):
After the changes are made, docs can be updated as well.

packages/core/src/lib/options.ts
```diff
@@ -31,6 +31,7 @@ export const resolveNetwork = (network: NetworkId | Network): Network => {
 
 export const resolveOptions = (params: WalletSelectorParams) => {
   const options: Options = {
+    languageCode: "",
```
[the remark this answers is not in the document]
HarisSQA: This is the default option, so if `options.languageCode == "" ` then it will choose the browser language.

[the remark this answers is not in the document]
kujtimprenkuSQA: This was addressed on this commit: https://github.com/near/wallet-selector/pull/728/commits/5ec4212ea0526c4af148e79e47f45112a98d7676

DamirSQA (approved):

PR #735, opened 2023-03-21
add: browser detect method for extension support

daoauth:
# Description

Some browsers can't use chrome extension, so add detection method for support browser.

It was writted by referring to [detect-browser](https://github.com/DamonOehlman/detect-browser/blob/master/src/index.ts), and test on 'safari', 'firefox', 'chrome', 'brave' and 'opera'.

@hcho112 

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings

### DIFF
--- a/packages/welldone-wallet/src/lib/welldone.ts
+++ b/packages/welldone-wallet/src/lib/welldone.ts
@@ -20,6 +20,7 @@ import type {
 import icon from "./icon";
 import { signTransactions } from "@near-wallet-selector/wallet-utils";
 import isMobile from "is-mobile";
+import { isSupportBrowser } from "./isSupportBrowser";
 
 export const STORAGE_ACCOUNT = "account";
 
@@ -360,8 +361,9 @@ export function setupWelldoneWallet({
   deprecated = false,
 }: WelldoneWalletParams = {}): WalletModuleFactory<InjectedWallet> {
   return async () => {
-    const mobile = isMobile();
-    if (mobile) {
+    const support =
+      !isMobile() && isSupportBrowser(["chrome", "edge-chromium", "opera"]);
+    if (!support) {
       return null;
     }
 
--- a/packages/welldone-wallet/src/lib/isSupportBrowser.ts
+++ b/packages/welldone-wallet/src/lib/isSupportBrowser.ts
@@ -1,0 +1,116 @@
+// https://github.com/DamonOehlman/detect-browser/blob/master/src/index.ts
+
+export type Browser =
+  | "aol"
+  | "edge"
+  | "edge-ios"
+  | "yandexbrowser"
+  | "kakaotalk"
+  | "samsung"
+  | "silk"
+  | "miui"
+  | "beaker"
+  | "edge-chromium"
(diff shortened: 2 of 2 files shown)


kujtimprenkuSQA (commented):
Hi, @daoauth thank you for submitting this issue.

packages/welldone-wallet/src/lib/welldone.ts
```diff
@@ -360,8 +361,9 @@ export function setupWelldoneWallet({
   deprecated = false,
 }: WelldoneWalletParams = {}): WalletModuleFactory<InjectedWallet> {
   return async () => {
-    const mobile = isMobile();
-    if (mobile) {
+    const support =
+      !isMobile() && isSupportBrowser(["chrome", "edge-chromium", "opera"]);
```
kujtimprenkuSQA, line 365: Why is it necessary to hide WELLDONE in other browsers?

We intentionally show them (injected wallets) in all browsers because it indicates that the `dApp` has support for that wallet, if you click the WELLDONE option in other browsers the modal will show you the `WalletNotInstalled` view.

hcho112: @kujtimprenkuSQA 

So far, it is not a good user experience when user sees a browser that doesn't support shows up and it redirects to a wrong/irrelevant page. (For example, Welldone wallet, if a user click them during using safari or firefox browser, it actually redirect user to a chrome extension download page. For the sake of simplify, we should only show wallets that are supported on that particular browser. (In case if any wallet teams want to support their wallet visible to many other browsers, they should work on plugin for particular browsers to support them natively)

Do you have different opinion to this? or alternative solution to above problem?

packages/welldone-wallet/src/lib/welldone.ts
```diff
@@ -360,8 +361,9 @@ export function setupWelldoneWallet({
   deprecated = false,
 }: WelldoneWalletParams = {}): WalletModuleFactory<InjectedWallet> {
   return async () => {
-    const mobile = isMobile();
-    if (mobile) {
+    const support =
```
hcho112, line 364: since this is boolean, lets re-name it to `isSupported`

hcho112 (approved):
LGTM, will consider merging this after approval from @kujtimprenkuSQA

packages/welldone-wallet/src/lib/welldone.ts
kujtimprenkuSQA: > Do you have different opinion to this? or alternative solution to above problem?

No, I don't have any other suggestion, based on what we discussed yesterday we will work on showing wallets only on the browsers they are supported on.

kujtimprenkuSQA:
@hcho112 LGTM too, your approval was dismissed due to the base branch change since we needed to follow the flow for contributions: https://github.com/near/wallet-selector/blob/main/CONTRIBUTING.md#branching .

Also had to update the description of the PR to include the **Type of change** checklist to make `gh-actions` pass.

kujtimprenkuSQA (approved):

PR #764, opened 2023-04-05
refactor code to remove sonarlint warning

sweetim:
there are 2 issues detected by sonarlint 
- missing dependency of `modal` in `useEffect`

image

- not using `useMemo` for context provider value

image

and added a loading state

### DIFF
--- a/examples/react/contexts/WalletSelectorContext.tsx
+++ b/examples/react/contexts/WalletSelectorContext.tsx
@@ -42,8 +42,8 @@ declare global {
 }
 
 interface WalletSelectorContextValue {
-  selector: WalletSelector;
-  modal: WalletSelectorModal;
+  selector: WalletSelector | null;
+  modal: WalletSelectorModal | null;
   accounts: Array<AccountState>;
   accountId: string | null;
 }
@@ -110,6 +110,8 @@ export const WalletSelectorContextProvider: React.FC<{
     const state = _selector.store.getState();
     setAccounts(state.accounts);
 
+    // this is added for debugging purpose only
+    // for more information (https://github.com/near/wallet-selector/pull/764#issuecomment-1498073367)
     window.selector = _selector;
     window.modal = _modal;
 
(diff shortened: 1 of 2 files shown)


sweetim:
and extra question, why `selector` and `modal` is exposed into the `window` variable?
is this being used anywhere?

```
window.selector = _selector;
window.modal = _modal;
```

examples/react/contexts/WalletSelectorContext.tsx
```diff
@@ -141,23 +143,25 @@ export const WalletSelectorContextProvider: React.FC<{
...
       onHideSubscription.remove();
     };
-  }, [selector]);
+  }, [selector, modal]);
 
-  if (!selector || !modal) {
+  const walletSelectorContextValue = useMemo<WalletSelectorContextValue>(
+    () => ({
+        selector: selector!,
+        modal: modal!,
```
sweetim, line 151: using `!` is not idle here, but i guess is ok since it will never pass undefined into the context provider, as it would have been returned from the loading condition, 

or i should change the interface to include `null` support?

kujtimprenkuSQA:
> and extra question, why `selector` and `modal` is exposed to the `window` variable? is this being used anywhere?
> 
> ```
> window.selector = _selector;
> window.modal = _modal;
> ```

It's not necessarily required to expose `selector` and `modal` into window for a dApp, the reason we expose them in our examples is that we can access them from the **Dev Tools Console** easily if we want to quickly check/test some functionality:

For example:

```ts
const wallet = await selector.wallet("sender");
const accounts = await wallet.signIn({ contractId: "test.testnet" });
```

examples/react/contexts/WalletSelectorContext.tsx
kujtimprenkuSQA: It's fine either way but I think it's best to change the`WalletSelectorContextValue`  interface.

examples/react/contexts/WalletSelectorContext.tsx
```diff
@@ -103,11 +110,14 @@ export const WalletSelectorContextProvider: React.FC<{
     const state = _selector.store.getState();
     setAccounts(state.accounts);
 
+    // this is added for debugging purpose only
```
sweetim, line 113: added comment here, so people are aware of it

kujtimprenkuSQA (approved):

PR #811, opened 2023-05-19
feat: Add `signInMulti` method

kujtimprenkuSQA:
# Description

This PR is about adding the `signInMulti` method to the public API of the wallets in Wallet Selector based on https://github.com/near/NEPs/pull/428 .

- Added interfaces in wallet.types for the `signInMulti`.
- Added `signInMulti` in each wallet, it initially throws when calling this method.
- Added logic in the `core` package to allow sign-in with multiple contracts.
- Updated `modal-ui` and `modal-ui-js` to pass the list of `contracts` to the `signInMulti` method.
- Added docs for `signInMulti`.

## Wallets that support `signInMulti` in this PR
- none

# Checklist:

# Type of change.


DamirSQA (dismissed):

amirsaran3 (dismissed):

packages/core/src/lib/services/wallet-modules/wallet-modules.service.ts
```diff
@@ -265,6 +298,25 @@ export class WalletModules {
         accounts,
         contractId,
         methodNames,
+        contracts: null,
+      });
+
+      return accounts;
+    };
+
+    wallet.signInMulti = async (params: never) => {
```
MaximusHaximus, line 307: Rather than having every wallet module be required by _convention_ to implement a `signInMulti()` method that `throw`'s an error if it isn't supported, can we instead _not_ add a `signInMulti()` method in the modules that do not support it, and make **this decorator** check if the method exists and throw the error
```
throw new Error(
        `The signInMulti method is not supported by ${metadata.name}`
      );
```

if it hasn't been implemented?

This has two benefits:
1. If any wallet module doesn't implement the method for whatever reason, callers will always get a consistent error thrown instead of there being a '_signInMulti() is not a function' type of error thrown from our attempt to call it in those cases
2. We don't need to inform wallet module authors about the need to add a stub, ensure their modules have the stub during code reviews, etc.  -- we can treat the lack of implementation as a 'not supported'.

packages/core/src/lib/store.types.ts
```diff
@@ -60,6 +62,10 @@ export interface WalletSelectorState {
    * Returns ID-s of 5 recently signed in wallets.
    */
   recentlySignedInWallets: Array<string>;
+  /**
+   * The list of contracts when SignInMulti is supported.
+   */
+  contracts: MultiContractState | null;
```
MaximusHaximus, line 68: We need to handle cases where:

1. A dapp might call `signIn()` for a contract that was previously signed into using `signInMulti()`.  For example, let's say that a dapp uses 3 contracts regularly, and their access key allowance expires for just _one_ of those three contracts.  The dapp author does not need to request the user re-signIn to all 3 contracts using `signInMulti()` -- they would just request `signIn()` for the _one contract_ whose allowance has been used up.
2. A dapp might call `signInMulti()` with an array that includes a contract that they had _previously_ signed in to using `signIn()`

If we always store results from `signInMulti()` in `contracts` and results from `signIn()` in `contract`, we will end up with duplicate entries in these keys when a mixture of calls is used by the dapp developer -- we need to have one-and-only-one entry for a given contract across these two fields (`contract` and `contracts`) to make sure everything is sane and deterministic.

MaximusHaximus (changes requested):
Added comments around keeping state for `signInMulti()` separate from `signIn()` -- we need to make sure that a combination of `signIn()` and `signInMulti()` being called with the same contracts will result in sane and deterministic behaviour.

packages/core/src/lib/services/wallet-modules/wallet-modules.service.ts
kujtimprenkuSQA: Addressed it in this commit: https://github.com/near/wallet-selector/pull/811/commits/8121e5e8f91ffc2ed6a71416a7a51a0db8cf1a95

packages/core/src/lib/store.types.ts
amirsaran3: This PR should solve these issues: #847

ailisp:
Hi @kujtimprenkuSQA @MaximusHaximus What is the status of this PR? Our team is recently tried to [add a relevant feature and discovered this](https://github.com/near/neardevhub-widgets/issues/137). I wonder if we can move this forward

kujtimprenkuSQA:
> Hi @kujtimprenkuSQA @MaximusHaximus What is the status of this PR? Our team is recently tried to [add a relevant feature](https://near.social/devgovgigs.near/widget/gigs-board.pages.Post?id=1033) and discovered this. [WIP branch](https://github.com/petersalomonsen/wallet-selector/blob/feat/multiple-local-keys/packages/my-near-wallet/src/lib/my-near-wallet.spec.ts#L256). I wonder if we can move this forward

Hi, this PR is not ready to be merged yet as we still need this additional PR https://github.com/near/wallet-selector/pull/847 related to the comment https://github.com/near/wallet-selector/pull/811#discussion_r1242130459 to be reviewed and make a decision if we should go with it or not since it will introduce breaking changes in Wallet Selector.

As of right now, we are not aware if any wallet supports the `signInMulti`.
For browser wallets I think this update in NAJ is needed: https://github.com/near/near-api-js/issues/1153 there was some work done related to it: https://github.com/near/near-api-js/pull/1163

petersalomonsen:
OK, so maybe I may move forward with my experiments in here then: https://github.com/near/wallet-selector/pull/887 ?

This approach does not require change in wallets, and does not introduce any breaking change to the wallet selector. It simply offers the option to add a function access key to another contract, and store it in a separate keystore ( near-api-js supports multiple keystore with different prefixes ). 

As you can see from the [test case](https://github.com/near/wallet-selector/pull/887/files#diff-430274eeefa33538a8818b0d157eb981fe8951390eb2ecf1b277fead00419963R256) there's a new method `addContractConnection` that you can call to add a function access key to an additional contract and store it locally.

For the particular issue for NEAR DevHub the idea is then to create the access key for the devhub contract when you submit something for the first time, and then reuse that access keys for later calls. This way, the user could also authenticate with other contracts on demand.

cc @ailisp @frol

petersalomonsen:
See a proof of concept demo video here ( for the wallet-selector guestbook ):

https://www.youtube.com/watch?v=ocZKP2-Lxes

ailisp:
@petersalomonsen the demo is very cool! I wonder how can we invoke `wallet.addContractConnection` from the DevHub widget, since there is no BOS api to access wallet directly

petersalomonsen:
> @petersalomonsen the demo is very cool! I wonder how can we invoke `wallet.addContractConnection` from the DevHub widget, since there is no BOS api to access wallet directly

@ailisp thanks, and actually you don't need to invoke that from the DevHub widget. This should be handled by the nearsocial viewer ( or maybe VM ) so that it detects if there's a wallet redirection, and then offers to add a function access key for that contract. Here's an example of how I've added this to the viewer for my PoC with the DevHub widget: https://github.com/NearSocial/viewer/pull/189/files

And this is also what's used in the demo you can see here: https://youtu.be/DVoZK5CJ2uE

petarvujovic98:
What is the plan for this feature moving forward? Any timeline estimates or prediction of when it could be available?

kujtimprenkuSQA:
> What is the plan for this feature moving forward? Any timeline estimates or prediction of when it could be available?

The initial work for adding the `signInMulti` into the wallet-selector has been done in this PR, but to address the comment left here https://github.com/near/wallet-selector/pull/811#discussion_r1242130459 there was a need to make some additional changes https://github.com/near/wallet-selector/pull/847 that would be breaking changes in wallet-selector.

We still need a review from Pagoda and further discuss this implementation before it's merged.

As mentioned in the above comments the currently available keystores will have to catch up in order to support multiple keys per account https://github.com/near/near-api-js/issues/1153 there was some work done related to it: https://github.com/near/near-api-js/pull/1163

PR #849, opened 2023-07-18
Metamask snap

AZbang:
I'm from the HERE Wallet team, which received a grant to develop and maintain the official Metamask plugin. We are happy to present our work!

## How it works:
The wallet-selector package uses the @near-snap/sdk library which interacts with near-api-js and the metamask snap rpc.

The @near-snap/plugin itself does not have internet access. The plugin only signs transactions and provides an interface for viewing actions signed by the user.

### Limited access to a specific contract.
Method near_connect(contractId, network, methods) works in the same way as limited keys, that is, they allow you to send free transactions without annoying confirmations.

### Details of what the user is signing.
List of transaction actions, confirmations of new permissions, additional information. Unfortunately at the moment snap-ui is not very rich, but we are doing our best to make the UI/UX of the plugin better for the user.

### Support for delegated transactions. 
And own infrastructure for new users. HERE Wallet allows metamask users (on the mainnet) to send gas free transactions without having funds on their account.

### Free account activation in mainnet and testnet.
For users, the HERE Wallet metamask automatically activates accounts so that later it will be possible to use delegated transactions to interact with dApps without having funds on the account. (In the future, we hope to involve the NEAR Foundation in this, as we believe this is an important part of onboarding new users to the network)

### Long term support from our team.
We have been developing the NEAR infrastructure for a year now and are not going to stop. We are very excited to contribute to another important ecosystem service.


DamirSQA:
Greetings, after some initial testing it seems that it is working properly. 

Some issues were noticed with the Metamask Flask. Install button is blocked/disabled, so in order to go through this step, flex property needed to be disabled with inspector as a workaround. 

no install
remove install

Aside from that, Near token amount is not being displayed after funding and importing Near token.  

image

DamirSQA:
@AZbang , all good, comments added are mainly for the future reference.
In that regard, here is[ the recording of the flow](https://www.loom.com/share/8e3b9ea584004955aac55f2c0a15ad2c). (transaction confirmation popup is not showing on recording though)

We have some questions about the packages:

- What’s the reason behind adding an additional package, instead replacing Near Snap?

- Why is Metamask name being used instead of Near, and what is idea behind the old Near Snap package, to deprecate it or to replace it? "

packages/metamask-snap/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "url": "https://github.com/near/wallet-selector.git"
+  },
+  "bugs": {
+    "url": "https://github.com/near/wallet-selector/issues"
+  },
+  "homepage": "https://github.com/near/wallet-selector/tree/main/packages/metamask-snap",
+  "peerDependencies": {
+    "near-api-js": "^1.0.0 || ^2.0.0"
+  },
+  "dependencies": {
```
DamirSQA, line 27: Dependencies needs to be added only in the root level `package.json`, and it will be detected automatcaly, so it should be removed from here.

packages/metamask-snap/yarn.lock
DamirSQA, line 1: Yarn lock inside metamask-snap needs to be removed.

examples/react/contexts/WalletSelectorContext.tsx
```diff
@@ -71,6 +72,7 @@ export const WalletSelectorContextProvider: React.FC<{
         setupMathWallet(),
         setupNightly(),
         setupMeteorWallet(),
+        setupMetamaskSnap({}),
```
DamirSQA, line 75: It shouldn't be required for empty object to be present. You can have a look on other wallets/packages.

DamirSQA (changes requested):
Added some comments, also there are linting issues that should be addressed.

AZbang:
@DamirSQA  I didn't know if the previous metamask plugin could be removed. Did this!

README.md
```diff
@@ -58,6 +58,7 @@ yarn add \
   @near-wallet-selector/sender \
   @near-wallet-selector/nearfi \
   @near-wallet-selector/here-wallet \
+  @near-wallet-selector/near-snap \
```
erditkurteshiSQA, line 61: Can you please remove this import because it's already imported?

README.md
```diff
@@ -81,6 +82,7 @@ npm install \
...
```
erditkurteshiSQA, line 85: Can you please remove this import because it's already imported?

examples/react/contexts/WalletSelectorContext.tsx
```diff
@@ -72,6 +72,7 @@ export const WalletSelectorContextProvider: React.FC<{
         setupNightly(),
         setupMeteorWallet(),
         setupNearSnap(),
+        setupNearSnap(),
```
erditkurteshiSQA, line 75: Can you please remove this import because it's already added to the line above?

package.json
```diff
@@ -64,6 +64,7 @@
     "build:wallet-utils": "nx run-many --target=build --projects=wallet-utils --configuration=production",
     "build:default-wallets": "nx run-many --target=build --projects=default-wallets --configuration=production",
     "build:near-snap": "nx run-many --target=build --projects=near-snap --configuration=production",
+    "build:near-snap": "nx run-many --target=build --projects=near-snap --configuration=production",
```
erditkurteshiSQA, line 67: Can you please remove this script, because it is duplicated?

packages/near-snap/README.md
```diff
@@ -1,7 +1,6 @@
 # @near-wallet-selector/near-snap
-[Metamask Snap](https://metamask.io/snaps/) system that allows anyone to safely expand the capabilities of MetaMask. Currently is pre-release software running on [Metamask Flask](https://metamask.io/flask/).
 
-This package implement NEAR snap for NEAR Wallet Selector.
+This is the NEAR Metamask Snap package for NEAR Wallet Selector.
```
erditkurteshiSQA, line 3: Can you please rename everything in this file to match with setupNearSnap?

packages/near-snap/project.json
```diff
@@ -51,5 +51,5 @@
       }
     }
   },
-  "tags": ["injected-wallet"]
+  "tags": ["browser-wallet"]
```
erditkurteshiSQA, line 54: Is there any particular reason why this is changed to browser-wallet?

packages/near-snap/tsconfig.json
```diff
@@ -3,9 +3,9 @@
   "compilerOptions": {
     "forceConsistentCasingInFileNames": true,
     "strict": true,
-    "noImplicitOverride": true,
     "noImplicitReturns": true,
-    "noFallthroughCasesInSwitch": true
+    "noFallthroughCasesInSwitch": true,
+    "module": "commonjs",
```
erditkurteshiSQA, line 8: In the project.json we already define the format. This is not needed.

tsconfig.base.json
```diff
@@ -33,35 +38,54 @@
       "@near-wallet-selector/here-wallet": [
         "packages/here-wallet/src/index.ts"
       ],
-      "@near-wallet-selector/ledger": ["packages/ledger/src/index.ts"],
+      "@near-wallet-selector/near-snap": [
+        "packages/near-snap/src/index.ts"
+      ],
```
erditkurteshiSQA, line 43: Can you please remove these lines, because they are duplicated?

erditkurteshiSQA:
Hi @AZbang ,

I left some comments regarding the last commit that needs to be addressed.
Also, these issues that are mentioned in this comment: https://github.com/near/wallet-selector/pull/849#issuecomment-1640464557 are still showing on some screens.

AZbang:
@erditkurteshiSQA  fixed all discussions! Also i added signMessage implementation

AZbang:
Also you can check mainnet guest-book demo 
https://near-snap.surge.sh/

DamirSQA:
Hi @AZbang it seems that Metamask(near) snap is not working properly after changes if not still work in progress, and Angular example seems to be broken, so it would be great if you can check that out as well.

AZbang:
@DamirSQA Can you find out more about what doesn't work? On the video, I check that everything works correctly:
https://www.veed.io/view/d7ff74f5-83ee-4d57-a0fb-7a38c6d64de3?sharingWidget=true&panel=share

The Angular example did indeed break, however this is not related to this PR because the problem is in the build system, I tried to fix it but I'm new to Angular and couldn't fix it quickly.

DamirSQA:
@AZbang it was the issue of wrong version in Metamask, after changing to v0.5.0 it worked properly.

DamirSQA:
@AZbang In order to fix Angular you can do the next:
- Install needed dependencies as devDependencies
`yarn add -D path-browserify`
`yarn add -D util`

- Update paths in `tsconfig.base.json` in wallet-selector (root) and add:
`"path": ["node_modules/path-browserify"]`

- This will now throw another error about the process version missing. 
To fix it update `polyfills.ts` in angular example add version to process:
`(window as any).process = { env: {}, version: "" };`

packages/near-snap/README.md
```diff
@@ -29,20 +28,22 @@ Then use it in your dApp:
...
-
 const selector = await setupWalletSelector({
   network: "testnet",
-  modules: [nearSnap],
+  modules: [setupNearSnap()],
 });
 ```
 
+
+## Here Wallet JS SDK
```
DamirSQA, line 38: Change Here Wallet to Near Snap

packages/near-snap/README.md
```diff
@@ -29,20 +28,22 @@ Then use it in your dApp:
...
+
+The library uses @near-snap/sdk, you can read more about the functionality here:
+https://github.com/here-wallet/near-snap
+
+
 ## Options
 
-- `iconUrl`: (`string?`): Image URL for the icon shown in the modal. This can also be a relative path or base64 encoded image. Defaults to `./assets/near-snap-icon.png`.
+- `iconUrl`: (`string?`): Icon is optional. Default image point to Here Wallet Logo in base64 format.
```
DamirSQA, line 46: Change Here Wallet to Near Snap

DamirSQA:
@AZbang  You can just do small changes in Readme, and we'll add Angular fix after merging.

DamirSQA (approved):

PR #883, opened 2023-08-10
feat: Add support for `signInMessage` method

kujtimprenkuSQA:
# Description

This PR is about adding support for the `signInMessage` method into the wallet-selector API.

This new method/function(way) will allow dApp users to "sign-in/login" into a dApp without a contractId (without adding LAK) by signing a message in the wallet based on the NEP413 standard for signing messages.

Once the message is signed and verified wallet selector will set the state as "signed-in"

Closes: https://github.com/near/wallet-selector/issues/921

### This PR includes

- Added interfaces for `signInMessage` method in core package and exposed them.
- Added support for signInMessage in the core package.
- Persisted the original `message` and the `signedInMessageAccount` into localStorage and rxjs store.
- Added a new `signInType()` function to the `selector` API.
- Added `signInMessage` support for all wallets that currently support `signMessage`.
- Added new helper function `verifyMessageNEP413` to core helper functions.
- Added a `signInMessage` method to the API of the modal, so we can open the modal in a state to handle `signInMessage` e.g `modal.signInMessage(params)`
- Updated examples, added a new button to trigger `signInMessage`
- Updated docs for the changes in this PR.

### Wallets that support `signInMessage` in this PR
- Meteor Wallet
- Here Wallet
- Near Snap
- Nightly Wallet
- WELLDONE Wallet
- NearMobileWallet
- MyNearWallet
- Sender

### How to use/trigger `signInMessage` in a dApp

> Note: All the examples below are documented in this PR.

#### Setup Wallet Selector & Modal

```ts
import { setupWalletSelector } from "@near-wallet-selector/core";
import { setupModal } from "@near-wallet-selector/modal-ui";
import { setupMeteorWallet } from "@near-wallet-selector/meteor-wallet";
import { setupHereWallet } from "@near-wallet-selector/here-wallet";

// Setup Wallet Selector
const selector = await setupWalletSelector({
  network: "testnet",
  modules: [
     setupMeteorWallet(),
     setupHereWallet(),
     // ...you can include other wallets too.
  ],
});

// Setup Modal
const modal = setupModal(selector, {
  contractId: "guest-book.testnet" // nothing changes here the `signIn` will use this contractId as always.
});

```
#### Trigger `signInMessage`
```ts

// Construct the message
const message = "test message for verification";
let nonceArray: Uint8Array = new Uint8Array(32);
nonceArray = crypto.getRandomValues(nonceArray);
const nonce = Buffer.from(nonceArray);
const recipient = "myapp.com";

[...]

PR #940, opened 2023-09-21
feat: Add nep-413 signMessage in WELLDONE Wallet

sunny0529:
# Description

Add nep-413 signMessage in WELLDONE Wallet.


packages/welldone-wallet/src/lib/welldone.ts
```diff
@@ -295,6 +331,15 @@ const WelldoneWallet: WalletBehaviourFactory<InjectedWallet> = async ({
...
     },
 
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    async signMessage(
+      message: Uint8Array | SignMessageParams,
+      accountId?: string
+    ) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      return (await signer.signMessage(message as any, accountId)) as any;
+    },
```
kujtimprenkuSQA, line 341: As mentioned in the comment: https://github.com/near/wallet-selector/issues/941#issuecomment-1730921716

This PR could be as simple as adding the signMessage this way, and to all the signing in the wallet

```ts
async signMessage({ message, nonce, recipient }) {
  return await _state.wallet.request("near", {
    method: "dapp:signMessage",
    params: [{ message, nonce, recipient }],
  });
},
```

The user does not need to be signed-in to wallet-selector to signMessage, this means dApp developers can directly call the `signMessage` without first calling the `signIn` method of wallet-selector.

```ts
const welldone = await selector.wallet("welldone-wallet");

await welldone.signMessage({
    message: "test message",
    nonce: [0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,24,26,27,28,29,30,31],
    recipient: "www.google.com",
});
```

kujtimprenkuSQA (changes requested):

kujtimprenkuSQA:
Hi, @sunny0529 just checking if there's any plan to address the comment, or if there's anything unclear about my suggestion.

packages/welldone-wallet/src/lib/welldone.ts
sunny0529: Thanks for the feedback. I committed the fix :)

Just double-checking because of the example you provided. Do you suggest NEP413 serializing inside the wallet? Instead we implemented it in the wallet-selector which is outside the wallet.

kujtimprenkuSQA: > Just double-checking because of the example you provided. Do you suggest NEP413 serializing inside the wallet? Instead we implemented it in the wallet-selector which is outside the wallet.

Yes, ideally it would be nice if the logic for serializing is kept in the wallet.

packages/welldone-wallet/src/lib/welldone.ts
```diff
@@ -295,6 +298,36 @@ const WelldoneWallet: WalletBehaviourFactory<InjectedWallet> = async ({
...
+      if (!_state.wallet) {
+        throw new Error("Wallet is not installed");
+      }
+
+      const accounts = getAccounts();
+      const account = accounts[0];
+
+      if (!account) {
+        throw new Error("Failed to find account for signing");
+      }
```
kujtimprenkuSQA, line 312: The way the `signMessage` should be working in wallet-selector:

A dApp should be able to trigger the `signMessage` of a wallet without first being signed-in.

```ts
const welldone = await selector.wallet("welldone-wallet");

//There should be no checks if a user is signed in or if there's an account.
const signedMessage = welldone.signMessage(params);
```
You could compare the flow of signMessage with other wallets like: Meteor, Nightly and Here Wallet:

```ts
const meteor = await selector.wallet("meteor-wallet");

// Prompts the user to sign the message without checking for the signed-in state.
const signedMessage = meteor .signMessage(params);
```

packages/welldone-wallet/src/lib/nep413.ts
```diff
@@ -43,16 +43,9 @@ export const payloadSchema = new Map([
...
-  return false;
-};
-
 export const serializeNep413 = (
   signMessageParams: SignMessageParams
 ): Buffer => {
   const payload = new Payload({ ...signMessageParams });
-  return Buffer.from(serialize(payloadSchema, payload));
+  return Buffer.from(serialize(payloadSchemaNep413, payload));
 };
```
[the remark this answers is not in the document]
kujtimprenkuSQA: Yes, I think that's a great idea, this would avoid potential human errors when others implement `signMessage` .

Can we expose it on this PR I think we could add it in this file:
https://github.com/near/wallet-selector/blob/main/packages/core/src/lib/helpers/verify-signature/payload.ts

packages/welldone-wallet/src/lib/welldone.ts
kujtimprenkuSQA: Understood, the serialization can be kept as it is now 👍 .

packages/welldone-wallet/src/lib/welldone.ts
kujtimprenkuSQA: I worry that this check will always throw unless we call `signIn` before it:

```ts
const accounts = getAccounts(); // returns [] unless we call `signIn`
const account = accounts[0];

if (!account) {
  throw new Error("Failed to find account for signing");
}
```

The `_state.account` gets assigned a value only inside the `signIn` method: https://github.com/near/wallet-selector/blob/main/packages/welldone-wallet/src/lib/welldone.ts#L245 therefore when checking for `getAccounts()` before `signIn` it will always return an `[]` empty array.

kujtimprenkuSQA: To reproduce it, first `signOut` from Welldone in the example and open the console in Chrome dev tools and try:

```ts
const welldone = await selector.wallet("welldone-wallet");

await welldone.signMessage({
    message: "test message",
    nonce: [0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,24,26,27,28,29,30,31],
    recipient: "www.some-website.com",
});
```

This would allow us to add support for Welldone in this feature too: https://github.com/near/wallet-selector/pull/883

sunny0529: Oh, I'm sorry. I had misunderstood our code.
You're right. Thanks to your kind explanation, I was able to fix the code :)

I modified the code to allow users and developers to call `signMessage` before `signIn`, thank you.

packages/welldone-wallet/src/lib/nep413.ts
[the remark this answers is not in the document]
sunny0529: Thank you for your positive consideration of my suggestion.

I've added `serializeNep413` in that file, please check it out.

packages/welldone-wallet/src/lib/welldone.ts
```diff
@@ -295,6 +301,45 @@ const WelldoneWallet: WalletBehaviourFactory<InjectedWallet> = async ({
...
+          }
+        : {
+            accountId,
+            publicKey: signed[0].publicKey,
+            signature: Buffer.from(
+              signed[0].signature.substr(2),
+              "hex"
+            ).toString("base64"),
+          };
+    },
```
kujtimprenkuSQA, line 341: This suggestion is more of a personal preference but I think this code would read better if we avoid the ternary operator in this case:

```ts
const result = {
  accountId,
  publicKey: signed[0].publicKey,
  signature: Buffer.from(signed[0].signature.substr(2), "hex").toString(
    "base64"
  ),
};

if (message.state) {
  return { ...result, state: message.state };
}

return result;

```

packages/welldone-wallet/src/lib/nep413.ts
```diff
@@ -0,0 +1,51 @@
+import type { SignMessageParams } from "@near-wallet-selector/core";
```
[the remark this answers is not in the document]
sunny0529: You're right. Thank you:)

packages/welldone-wallet/src/lib/welldone.ts
sunny0529: Thank you for the feedback.

kujtimprenkuSQA:
The PR looks good to me now, @sunny0529 thank you for submitting this PR and addressing the comments.

kujtimprenkuSQA (approved):

PR #1003, opened 2023-11-07
modify signmessage payload schema nonce type

AmazingBeerBelly:
# Description

Modify SignMessage PayloadSchema's nonce type from [32] to ["u8"]

# Checklist:

- [ ] My changes generate no new warnings

# Type of change.

### DIFF
--- a/packages/core/src/lib/helpers/verify-signature/payload.ts
+++ b/packages/core/src/lib/helpers/verify-signature/payload.ts
@@ -29,7 +29,7 @@ export const payloadSchema = new Map([
       fields: [
         ["tag", "u32"],
         ["message", "string"],
-        ["nonce", [32]],
+        ["nonce", ["u8"]],
         ["recipient", "string"],
         [
           "callbackUrl",


packages/core/src/lib/helpers/verify-signature/payload.ts
```diff
@@ -29,7 +29,7 @@ export const payloadSchema = new Map([
       fields: [
         ["tag", "u32"],
         ["message", "string"],
-        ["nonce", [32]],
+        ["nonce", ["u8"]],
```
kujtimprenkuSQA, line 32: Why is this needed? - this will break all other implementations.

kujtimprenkuSQA: The NEP413 states: 
>A nonce that uniquely identifies this instance of the message, denoted as a 32 bytes array (a fixed `Buffer` in JS/TS).

From my understanding in borsh-js `v0.7.0` an array of unsigned 8-bit integers with a fixed length of 32 elements is done as we already have `["nonce", [32]]`

By, changing the type to `["nonce", ["u8"]]` would allow any length of 8-bit integer elements.

AmazingBeerBelly: Ok, it's my fault. I using v1.0.0, and set struct to "nonce: { array: { type: 'u8', len: 32 } }" will work.

Close this PR.

kujtimprenkuSQA: I see, in v1.0.0 borsh has been re-written, you probably need to do something like this:

```ts
const schema = {
    struct: {
        tag: 'u32',
        message: 'string',
        nonce: { array: { type: 'u8', len: 32 } },
        recipient: 'string',
        callbackUrl: { option: 'string' },
    },
};
```

PR #1011, opened 2023-11-17
Example with Vanilla JS

petersalomonsen:
# NEAR wallet selector - vanilla JS example

This example shows how to use the NEAR wallet selector from a "vanilla" Javascript app, without using any framework like React or Angular.

In order to handle the imports and [importmap](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap) is used. This has to be re-generated each time there is a change in the imports, but otherwise you can just make changes to the javascript or html and it will be visible instantly. For generating the import map [JSPM](https://jspm.org/) is used. The benefit of this while developing is that there is no re-bundling time like when using e.g. WebPack with React or Angular.

JSPM can also be used for creating a single file production bundle if needed.

## Creating the import map

There's a yarn script for creating the importmap:

`yarn examples:vanillajs:generateimportmap`

## Serving the static files

[http-server](https://github.com/http-party/http-server) is used for serving the static files if you run:

`yarn examples:vanillajs:serve`


examples/vanillajs/main.js
```diff
@@ -0,0 +1,15 @@
...
+
+const selector = await setupWalletSelector({
+  network: "testnet",
+  modules: [setupMyNearWallet(), setupHereWallet()],
+});
+
+const modal = setupModal(selector, {
+  contractId: "test.testnet",
+});
+
```
kujtimprenkuSQA, line 14: Could you add the `selector` and `modal` into the global `window` object?

```ts
window.selector = selector;
window.modal = modal;
```

Since the html document is pretty empty we can at least interact with the wallet selector through DevTools of the browser.

package.json
```diff
@@ -64,6 +64,8 @@
     "build:account-export": "nx run-many --target=build --projects=account-export --configuration=production",
     "build:ramper-wallet": "nx run-many --target=build --projects=ramper-wallet --configuration=production",
     "build:near-mobile-wallet": "nx run-many --target=build --projects=near-mobile-wallet --configuration=production",
+    "examples:vanillajs:generateimportmap": "(cd examples/vanillajs && jspm link index.html -o index.html)",
+    "examples:vanillajs:serve": "http-server examples/vanillajs",
```
kujtimprenkuSQA, line 68: Would it make sense to combine these two scripts (pretty difficult to remember them)?

Also, let's rename `examples:vanillajs:serve` to `serve:vanillajs` and move it down to the other scripts for serving angular and react examples.

I would suggest that we disable cache for this example by adding the option `-c-1` to the serve script because the browser is immediately caching `main.js` and changes are not being picked without hard-reload.

kujtimprenkuSQA (commented):
Hi, @petersalomonsen thank you for submitting this PR.

Can you please change the base branch from `main` to `dev`?

petersalomonsen:
> Hi, @petersalomonsen thank you for submitting this PR.
> 
> Can you please change the base branch from `main` to `dev`? https://github.com/near/wallet-selector/blob/main/CONTRIBUTING.md#branching

I have done this now, sorry for not catching this immediately.

kujtimprenkuSQA:
This looks good to me, could you just merge your branch with the latest changes of `dev` and fix the conflicts? - if you have any difficulties let me know.

kujtimprenkuSQA (approved):

PR #1029, opened 2023-12-05
fix: Update Meteor Wallet SDK

lostpebble:
We have updated the Near API JS library in our SDK, we think this was the main cause for recent signing issues- most likely a mismatch between the imports and serialization issues.

Along with this, we've also moved more of the logic to within the SDK itself instead of the Wallet Selector package.


kujtimprenkuSQA (approved):
Thank you for submitting this PR.

PR #1077, opened 2024-02-13
feat: Update HERE Wallet  and add topLevelInjected

AZbang:
It was necessary to add the ability to associate an application inside an iframe with a parent window and allow seamless initialization and method calls. This will help integrate applications inside wallets. For example Telegram HERE Wallet:

https://github.com/near/wallet-selector/assets/14349805/0e9bbc9f-5222-40bf-9592-3689e151d317

`topLevelInjected` method is needed because wallet-selector is designed in such a way that only modal-ui knows about the contract that the application uses.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "@angular/platform-browser": "15.2.9",
     "@angular/platform-browser-dynamic": "15.2.9",
     "@angular/router": "15.2.9",
-    "@here-wallet/core": "^1.5.1",
+    "@here-wallet/core": "^1.6.6",
     "@jscutlery/semver": "3.1.0",
     "@ledgerhq/hw-transport": "6.30.3",
     "@ledgerhq/hw-transport-webhid": "6.28.3",
--- a/packages/core/docs/api/wallet.md
+++ b/packages/core/docs/api/wallet.md
@@ -63,7 +63,16 @@ There are four wallet types:
 
 **Description**
 
-Returns meta information about the wallet such as `name`, `description`, `iconUrl` , `deprecated` and `available` but can include wallet-specific properties such as `downloadUrl` and `useUrlAccountImport` for injected wallets or `contractId`,  `runOnStartup` for instant-link wallets and `walletUrl` for browser wallets.
+Returns meta information about the wallet such as `name`, `description`, `iconUrl` , `deprecated` and `available` but can include wallet-specific properties such as `downloadUrl`, `useUrlAccountImport` and `topLevelInjected` for injected wallets or `contractId`,  `runOnStartup` for instant-link wallets and `walletUrl` for browser wallets.
+
+- `name`: Displayed in modal-ui as wallet name
+- `description`: Displayed in modal-ui as wallet description
+- `iconUrl`: Displayed in modal-ui as wallet icon
+- `deprecated`: Makes the wallet unselectable via modal-ui
+- `available`: Makes the wallet unselectable via modal-ui, use if the wallet cannot be selected in the user's environment.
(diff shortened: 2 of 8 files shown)


ewiner:
@AZbang can you help me understand why `topLevelInjected` is different from a normal injected wallet?  (Also if we go forward with this PR, you'll have to document that option anyway.)

AZbang:
@ewiner  Added a description to the metadata parameters

AZbang:
@ewiner The video in the PR description shows that when opening a site inside an iframe, here-wallet transmits wallet data inside the site through here-wallet/core using an iframe postMessage. However, in order for the wallet to be immediately initialized inside wallet-selector, it is necessary to call initialization through modal-ui. The only way to tell modal-ui whether to signIn immediately is to pass a special parameter to metadata.

`TopLevel` means that the wallet is injected from the parent window or, in the case of native mobile applications, the WebView controller.

ewiner:
So the idea is for an injected wallet to force automatic sign-in, so the user doesn't have to click "Connect Wallet" or "Sign In".

I'm new to this codebase, but two questions/comments come to mind:
1. Shouldn't that be a decision for the app developer to make instead of the wallet provider?  Are there apps that would break if the signIn happens upon initialization instead of when the user asks?
2. The wallet selector supports multiple injected wallets - it seems to me like with this change one of those wallets could force sign-in and prevent you from using others.

AZbang:
@ewiner 
1) wallet-selector remembers the last authorized wallet between sessions and automatically logs in upon initialization; the application should not care how wallet-selector received the active wallet in own reactive state. By design of the library, the application must subscribe to changes in the selector state.
Adding the same code to every application seems less sensible than simply updating a tool that is designed to provide access to wallets through a unified API.

2) In general, yes, if there are two topLevelInjected wallets, then wallet-selector will consider the last one in the list to be the active wallet.
However, in practice, topLevelInjected should only be used if the wallet is confident that it runs in its native environment and will not conflict. For example, native WebView controllers or embedded widgets on websites.

trechriron:
This PR will be eligible for merge into main after review by community and Pagoda. Thanks!

trechriron:
@kujtimprenkuSQA is no longer working on this project. If he wants to volunteer his personal time, it would be welcome, however I don't want any SQA folks added as reviewers. I am trying to encourage the community to review this PR. This is still in progress. Thanks!

hcho112:
Agree with @ewiner. This PR change is tailored for specific problem that Here wallet is experiencing. Not disagree with the direction of the PR, but ideally this should have been [proposed on NEP](https://github.com/near/NEPs) and approved by the community and users. 

Or at least, I would prefer to get reviewed/feedback by other builder group community, specially around the maintainers of other injected wallets.

pvolnov:
The idea of wallet-selector is to help projects have access to users. Now we have a Telegram Wallet where 95% of all active users are on NEAR. At the same time, no project on NEAR can provide services to such users. By delaying, you are isolating the ecosystem from users. 

I believe that taking into account the real state of affairs in the ecosystem, this PR is for the public good and should be implemented in the near future.

AZbang:
@ewiner @trechriron @hcho112 
I really don't understand why you consider topLevelInjected a specific feature for HERE Wallet. Any other wallets will be able to use this to integrate web dApps inside their mobile wallets. This seems very useful. Just because we're the only ones who want to use it doesn't mean it's a feature only for us.

Patrick1904 (approved):
LGTM

vikpande:
thank you @Patrick1904

PR #1182, opened 2024-08-27
fix FailoverRpcProvider import

kdonthi:
Thanks for taking the time to submit a pull request!

If you haven't already, be sure to take a look at the [CONTRIBUTING](https://github.com/near/wallet-selector/blob/main/CONTRIBUTING.md) documentation to get up to speed with the practices in this repository.

When you're ready, switch to the "Preview" tab to select an applicable template:

- [General template (most common)](?expand=1&template=general.md)
- [Wallet integration template](?expand=1&template=wallet_integration.md)

### DIFF
--- a/packages/core/src/lib/services/provider/provider.service.ts
+++ b/packages/core/src/lib/services/provider/provider.service.ts
@@ -12,12 +12,13 @@ import type {
 } from "./provider.service.types";
 import { JsonRpcProvider } from "near-api-js/lib/providers";
 import type { SignedTransaction } from "near-api-js/lib/transaction";
+import {FailoverRpcProvider} from "@near-js/providers";
 
 export class Provider implements ProviderService {
   private provider: nearAPI.providers.FailoverRpcProvider;
 
   constructor(urls: Array<string>) {
-    this.provider = new nearAPI.providers.FailoverRpcProvider(
+    this.provider = new FailoverRpcProvider(
       this.urlsToProviders(urls)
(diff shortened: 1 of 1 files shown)


andy-haynes:
What problem does this address? That class is just re-exported from `near-api-js`, so this change should be a no-op.

Importing it like this only adds a dependency reference without adding it to `package.json`.

kdonthi:
> What problem does this address? That class is just re-exported from `near-api-js`, so this change should be a no-op.
> 
> Importing it like this only adds a dependency reference without adding it to `package.json`.

I keep getting this error whenever I start the multichain example and I end up just having to manually make that change in the node_modules.

`TypeError: nearAPI.providers.FailoverRpcProvider is not a constructor
    at new Provider (index.js:41:21)
    at index.js:3054:15
    at Generator.next (<anonymous>)
    at fulfilled (index.js:27:58)`

andy-haynes:
My guess is this happens because `nearAPI` is referencing an older version of `near-api-js` (one which doesn't export `FailoverRpcProvider`) but somehow  there is a more recent version of `@near-js/providers` installed as a secondary dependency. So it's important to determine where the issue originates because a fix like this, while convenient, further compounds the dependency issue.

Is there a GitHub issue you could link to this PR where you describe the problem in more detail? Would you mind please creating one if it does not exist and link it to this PR? I'm not sure how to reproduce this from your description.

PR #1199, opened 2024-09-10
Add fr and el translation

rainbowteam002:
Add fr and el translation


gagdiez:
ChatGPT says the greek translation is correct, and the FR is good enough according to me

gagdiez (approved):

trechriron (approved):

PR #1200, opened 2024-09-10
Add it and hu translations

citadeloneteam002:
Add it and hu translations #1200


gagdiez:
chatGPT says the hungarian one is correct, I can vouch for the italian one being right

gagdiez (approved):

trechriron (approved):

PR #1213, opened 2024-10-01
feat: Add signTransaction() and signDelegateAction() to BaseWalletBehaviour interface

gtsonevv:
# Description

# Checklist:

# Type of change.

### DIFF
--- a/packages/core/src/lib/wallet/wallet.types.ts
+++ b/packages/core/src/lib/wallet/wallet.types.ts
@@ -146,7 +146,7 @@ interface SignTransactionParams {
 export interface SignDelegateActionParams {
   blockHeightTtl: number;
   receiverId: string;
-  actions: Array<DelegateAction>;
+  actions: Array<Action>;
   callbackUrl?: string;
 }
 
--- a/packages/core/src/lib/wallet/wallet.types.ts
+++ b/packages/core/src/lib/wallet/wallet.types.ts
@@ -13,8 +13,10 @@ import type { FinalExecutionOutcome } from "near-api-js/lib/providers";
 import type {
   SignedTransaction,
   Transaction as Tx,
+  Signature,
 } from "near-api-js/lib/transaction";
 import type { Signer } from "near-api-js/lib/signer";
+import type { PublicKey } from "near-api-js/lib/utils";
 
 interface BaseWalletMetadata {
   /**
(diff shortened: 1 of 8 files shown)


packages/core/src/lib/wallet/wallet.types.ts
```diff
@@ -121,6 +128,112 @@ interface SignAndSendTransactionsParams {
...
 
+interface SignTransactionParams {
+  /**
+   * The Transaction object to sign
+   */
+  transaction: Tx;
+  /**
+   * The {Signer} object that assists with signing keys
+   */
+  signer: Signer;
```
frol, line 139: How do you envision the use of it? Isn't the whole point to use an external wallet as signer? How is it required?

packages/core/src/lib/wallet/wallet.types.ts
```diff
@@ -121,6 +128,112 @@ interface SignAndSendTransactionsParams {
   transactions: Array<Optional<Transaction, "signerId">>;
 }
 
+interface SignTransactionParams {
+  /**
+   * The Transaction object to sign
+   */
+  transaction: Tx;
```
frol, line 135: App developer in most of the cases cannot construct the full transaction since there is no public key and therefore there is no nonce. Also, it is often no account id either.

I cannot think of the use for `SignTransactionParams` interface as it is defined right now.

packages/core/src/lib/wallet/wallet.types.ts
```diff
@@ -121,6 +128,112 @@ interface SignAndSendTransactionsParams {
...
+
+interface SignTransactionActionsParams {
+  /**
+   * The NEAR account ID of the transaction receiver.
+   */
+  receiverId: string;
+  /**
+   * Tx nonce.
+   */
+  nonce: bigint;
```
frol, line 158: Tx nonce is attached to the public key nonce of the account, it does not make sense to require it from the app developer.

packages/core/src/lib/wallet/wallet.types.ts
```diff
@@ -121,6 +128,112 @@ interface SignAndSendTransactionsParams {
...
+  /**
+   * The maximum block height for which this action can be executed as part of a transaction
+   */
+  maxBlockHeight: bigint;
+  /**
+   * Public key for the access key used to sign the delegate action
+   */
+  publicKey: PublicKey;
```
frol, line 213: App developer does not have the user's public key

packages/core/src/lib/wallet/wallet.types.ts
```diff
@@ -121,6 +128,112 @@ interface SignAndSendTransactionsParams {
...
+
+interface MessageSigner {
+  sign(message: Uint8Array): Promise<Uint8Array>;
+}
+
+interface DelegateAction {
+  /**
+   * Account ID for the intended signer of the delegate action
+   */
+  senderId: string;
```
frol, line 189: App developer does not know the senderId

packages/core/src/lib/wallet/wallet.types.ts
```diff
@@ -121,6 +128,112 @@ interface SignAndSendTransactionsParams {
...
+   */
+  blockHeightTtl: number;
+  /**
+   * Account ID for the intended receiver of the meta transaction
+   */
+  receiverId: string;
+  /**
+   * Current nonce on the access key used to sign the delegate action
+   */
+  nonce: bigint;
```
frol, line 205: Same problem as with transactions, see above

frol (changes requested):
So far it looks totally pointless. Do you have a proof of concept? We need a dApp with these interfaces used and miniwallet that will handle these operations.

What you also should think about is how dApp will receive the signed transaction from:

1. web wallet - given that currently transaction signing experience redirects dApp user to a different website which redirects them back, so Promise is not going to work well here - we need to consider if we can open a new window instead of a redirect, and how that window will report the signed transaction back

2. mobile wallet on the same device (when Wallet Selector-enabled website is used on the mobile device where I have the wallet app) - it will open a separate app, which somehow should give the control back

3. mobile wallet for a desktop dApp - users are asked to scan a QR code, but again dApp needs to get a response back somehow

4. browser extension

5. Ledger device

trechriron:
@gtsonevv  - this PR was skipped for release 8.10.0 because changes were not addressed.

PR #1293, opened 2025-02-06
Bring back Hot Wallet & fix lint issue

gagdiez:
This PR brings back hot wallet after we reverted it from `main` and tried to add it in `dev` again


trechriron (approved):
Amazing!

PR #1323, opened 2025-04-16
fix: metamask onboarding

lampamazaza:
## Description
This pr introduces a fix for MetaMask wallet. It fixes the onboarding of new users

Metamask has a [check](https://github.com/MetaMask/core/blob/v360.0.0/packages/transaction-controller/src/utils/validation.ts#L100-L110) that prevents the execution of 'external' transactions (i.e., those originating outside MetaMask), when these transactions have an 'internal' account address in the to field (i.e., a user's own address within MetaMask) and contain non-empty data. When issuing an onboarding transaction, we set the to field to the user's own address and utilize the data field to pass required parameters to the Wallet Contract, specifically to add the provided public key to the account.
      
      
We set hash to value for AddKey/DeleteKey to bypass that metamask check. In the Wallet Contract itself contract compares to address with address hash.

### DIFF
--- a/packages/ethereum-wallets/src/lib/index.ts
+++ b/packages/ethereum-wallets/src/lib/index.ts
@@ -168,13 +168,15 @@ const EthereumWallets: WalletBehaviourFactory<
     tx: Transaction;
     relayerPublicKey: string;
   }): Promise<`0x${string}`> => {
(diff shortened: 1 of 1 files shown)


gagdiez:
@trechriron maybe we can release this as a fix for `8` and also include it automatically in `9`?

gagdiez (approved):

trechriron (approved):
Solid Fix. Thanks for the comments!

PR #1345, opened 2025-06-02
feat: Implement createSignedTransaction and returns SignedTransaction

ahnafalfariza:
# Description
This pull request introduces the new `createSignedTransaction` and `signTransaction` methods to the wallet-selector. This method enables wallets to create signed transactions and return a Signed Transaction object, empowering developers to handle raw signed transactions for custom workflows.

## Changes
- Added the `createSignedTransaction(receiverId, actions)` method and `signTransaction(transaction)` method to the wallet type and behavior interface.
- Added mandatory `signTransaction` as a mandatory method for each wallet
- Implemented or stubbed `createSignedTransaction` in all wallet modules. For unsupported wallets, the method includes logging parameters and throwing "not supported" errors.
- Extended wallet type definitions and documentation to include signing transaction.
- React example app adjustments:
  - Added a "Sign only and Return Signature" checkbox to the form.
  - Submission handler uses `createSignedTransaction` if the checkbox is checked, showing the signature in an alert and logging it to the console.

## Motivation

Previously, wallet-selector did not provide a unified way to create and retrieve signed transactions without immediately sending them. This new method improves flexibility for dApp developers who need to inspect, store, or relay signed transactions directly.

## Usage Example

```typescript
// Example usage in a dApp:
const signedTx = await wallet.createSignedTransaction(receiverId, actions);
console.log(signedTx); // Outputs the Signed Transaction object
```

In the React example app, select "Sign only and Return Signature" to try this flow.

## Additional Notes

- Not all wallets currently support this method; unsupported wallets will throw a clear error.
- These changes are backwards-compatible and do not affect existing transaction flows.

## Reference

- The design and implementation of `createSignedTransaction` [`near-api-js` Account.signTransaction](https://github.com/near/near-api-js/blob/master/packages/accounts/src/account.ts#L270-L273).

# Checklist:

# Type of change.

### DIFF
--- a/examples/react/components/Content.tsx
+++ b/examples/react/components/Content.tsx
@@ -129,7 +129,8 @@ const Content: React.FC = () => {
       : "Failed to verify";
 
     alert(
-      `${alertMessage} signed message: '${message.message
+      `${alertMessage} signed message: '${
+        message.message
       }': \n ${JSON.stringify(signedMessage)}`
     );
   };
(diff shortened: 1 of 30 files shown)


gagdiez:
hi @ahnafalfariza, thanks for the proposal! 

Currently we are aiming to make `wallet-selector` compatible with `near-api-js`, ideally, anything you built something using `naj` you can easily plug `wallet-selector` and vice versa

This means that we need `naj` and `wallet-selector` to share a common interface

Currently, there are two relevant classes in `naj`: `Account` and `Signer`

- `Account` implements [`createSignedTransaction(receiverId: string, actions: Action[] )`](https://github.com/near/near-api-js/blob/master/packages/accounts/src/account.ts#L270-L273)
- `Signer` implements [`signTransaction(tx: Transaction)`](https://github.com/near/near-api-js/blob/master/packages/signers/src/signer.ts#L57-L59)

I would strongly recommend to stick with one of those interfaces (in this case,  I guess you want to rename your function to `createSignedTransaction` and make sure the interfaces are compatible), so people can in the future build libraries that are agnostic to what library is used to be signing transactions

gagdiez (changes requested):
hi @ahnafalfariza, thanks for the proposal! 

Currently we are aiming to make `wallet-selector` compatible with `near-api-js`, ideally, anything built something using `naj` you can easily switch to `wallet-selector` and vice versa

This means that we need `naj` and `wallet-selector` to share a common interface

Currently, there are two relevant classes in `naj`: `Account` and `Signer`

- `Account` implements [`createSignedTransaction(receiverId: string, actions: Action[])`](https://github.com/near/near-api-js/blob/master/packages/accounts/src/account.ts#L270-L273)
- `Signer` implements [`signTransaction(tx: Transaction)`](https://github.com/near/near-api-js/blob/master/packages/signers/src/signer.ts#L57-L59)

I would strongly recommend to stick with one of those interfaces (in this case,  I guess you want to rename your function to `createSignedTransaction` and make sure the interfaces are compatible), so people can in the future build libraries that are agnostic to _what_ is signing transactions

roman009:
thx @gagdiez. makes sense
we want to get this functionality out asap as we have projects waiting for it. but we'll consider your suggestion

gagdiez (approved):

gagdiez:
ok, I'll not block it, but let me know if you cannot take on this, I'll ask somebody on my team to take a look

gagdiez:
lgtm

roman009 (approved):

Issue #1346, opened 2025-06-02
[Defect] Meteor Wallet Function Call Key Failing


gagdiez:
**Describe the bug**

In the frontend example, if you create a function call key with Meteor Wallet an error will raise:

```
signerId is undefined
```

I suspect the error comes from some incompatibility between the `Account` being created to sign the transaction in wallet selector, and the `Account` being used within the `meteor-wallet-sdk` 

**Steps To Reproduce**å

_Steps to reproduce the behavior:_

1. Run the example (`yarn serve:react`)
2. Modify the example so it creates a function call key (i.e. use the `createKeyFor`) option when initializing the selector
3. Try to sign a message into the guestbook

PR #1347, opened 2025-06-02
Force use of near-api-js@5.0.0 to fix signerId is undefined

matiasbenary:
Based on the issue reported in https://github.com/near/wallet-selector/issues/1346,  where `signerId is undefined` when using Meteor Wallet to create a function call key, I've pinned `near-api-js` to version 5.0.0 to resolve the incompatibility.

This version has been tested and works correctly .

This image shows a test performed on a HOT wallet
Captura desde 2025-06-02 14-18-55

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "prepack": "yarn build:core && yarn build:all",
     "prepublishOnly": "yarn build:core && yarn build:all",
     "serve:angular": "nx serve angular",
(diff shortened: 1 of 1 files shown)


matiasbenary:
I have this error with 5.1.1
image

ahnafalfariza:
Ah, deleting yarn.lock worked, thank you! I initially thought it might be a caching issue on my end.
Tested the fix and it's all good from me! 👍

PR #1361, opened 2025-07-08
Implement `Signer` interface for core `Wallet` and adapters

denbite:
Thanks for taking the time to submit a pull request!

If you haven't already, be sure to take a look at the [CONTRIBUTING](https://github.com/near/wallet-selector/blob/main/CONTRIBUTING.md) documentation to get up to speed with the practices in this repository.

When you're ready, switch to the "Preview" tab to select an applicable template:

- [General template (most common)](?expand=1&template=general.md)
- [Wallet integration template](?expand=1&template=wallet_integration.md)

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,7 @@
     "@ledgerhq/hw-transport": "6.30.3",
     "@ledgerhq/hw-transport-webhid": "6.29.4",
     "@meteorwallet/sdk": "1.0.9",
+    "@near-js/signers": "2.1.0",
     "@near-snap/sdk": "0.6.0",
     "@noble/hashes": "1.7.1",
     "@peersyst/near-mobile-signer": "1.1.4",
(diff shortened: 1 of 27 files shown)


ahnafalfariza:
since `getAccounts` also returns a public key, could `getPublicKey` simply return that value? I was wondering if so, does that mean we can use the same code?

denbite:
@ahnafalfariza the exact same idea has crossed my mind as well
however, most wallets don't return a valid `publicKey` in `getAccounts`, I've tested it myself - some return an empty `string` or `undefined`, and in the worst case, even a random key
so, this method isn't reliable at the moment, and if it ever starts returning real data, I'd be happy to change the logic so `getPublicKey` would simply parse public key from the response of `getAccounts`

ahnafalfariza:
got it, thanks for clarification!

ahnafalfariza (approved):

PR #1405, opened 2025-08-27
Naj Compatibility

gagdiez:
## Pre-flight checklist

- [x] I have read the [Contributing Guidelines on pull requests](https://github.com/near/wallet-selector/blob/main/CONTRIBUTING.md).
- [x] Commit messages follow the [conventional commits](https://www.conventionalcommits.org/) spec
- [x] **If this is a code change**: I have written unit tests.
- [x] **If this changes code in a published package**: I have run `pnpm changeset` to create a `changeset` JSON document appropriate for this change.
- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

This PR makes it possible to use the wallet selector as a `near-api-js` `Signer` by introducing a helper function that transforms `naj` `Actions` into the internal `Action`, and decorate the `signAndSendTransaction(s)` functions to use them

### DIFF
--- a/examples/react/wallets/web3modal.ts
+++ b/examples/react/wallets/web3modal.ts
@@ -69,5 +69,3 @@ export const web3Modal = createAppKit({
   coinbasePreference: "eoaOnly", // Smart accounts (Safe contract) not available on NEAR Protocol, only EOA.
   allWallets: "SHOW",
 });
-
-reconnect(wagmiAdapter.wagmiConfig);
--- a/packages/react-hook/src/lib/WalletSelectorProvider.tsx
+++ b/packages/react-hook/src/lib/WalletSelectorProvider.tsx
@@ -68,7 +68,7 @@ export interface WalletSelectorProviderValue {
   getAccessKeys: (accountId: string) => Promise<Array<unknown>>;
   signAndSendTransactions: (params: {
     transactions: Array<Transaction>;
-  }) => Promise<Array<object | string | number | null>>;
+  }) => Promise<Array<FinalExecutionOutcome>>;
   signMessage: (params: SignMessageParams) => Promise<void | SignedMessage>;
   getAccount: (accountId: string) => Promise<QueryResponseKindWithAmount>;
   verifyMessage: (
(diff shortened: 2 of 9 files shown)


examples/react/wallets/web3modal.ts
```diff
@@ -69,5 +69,3 @@ export const web3Modal = createAppKit({
   coinbasePreference: "eoaOnly", // Smart accounts (Safe contract) not available on NEAR Protocol, only EOA.
   allWallets: "SHOW",
 });
-
-reconnect(wagmiAdapter.wagmiConfig);
```
gagdiez, line 73: `ethereum wallets` need to be fixed (the "reconnect on reload page" is not working)

packages/react-hook/src/lib/WalletSelectorProvider.tsx
```diff
@@ -68,7 +68,7 @@ export interface WalletSelectorProviderValue {
   getAccessKeys: (accountId: string) => Promise<Array<unknown>>;
   signAndSendTransactions: (params: {
     transactions: Array<Transaction>;
-  }) => Promise<Array<object | string | number | null>>;
```
gagdiez, line 71: Hooks were using the wrong interface!

packages/wallet-utils/src/lib/create-action.ts
```diff
@@ -16,11 +16,7 @@ const getAccessKey = (permission: AddKeyPermission) => {
 };
 
 // TODO: Remove this function after all wallets use the NAJ Action by default
-export const createAction = (action: Action | NAJAction): NAJAction => {
```
gagdiez, line 19: not needed anymore, as now all inputs should be internal `Actions`

packages/wallet-utils/src/lib/sign-transactions.ts
```diff
@@ -1,4 +1,5 @@
 import type { Signer } from "near-api-js";
+
```
gagdiez, line 2:
```suggestion
```

ahnafalfariza:
yeah, I agree, I think this is fine for now. I was actually looking into this yesterday, replacing the interface with NAJ actions would likely introduce significant changes and require a major version bump due to the interface updates. I’m also not fully sure yet whether the wallet supports both internal actions and NAJ actions simultaneously

with the current changes, though, it looks like dapps will have more flexibility: dapps can use either internal actions or NAJ actions, and over time we can gradually migrate everything fully to NAJ action. what do you think?

ahnafalfariza (approved):

gagdiez:
yeah, for now this is fine, but we will need that mayor release eventually

which is fine, because we need compatibility with naj

Issue #1400, opened 2025-08-25
Use `near-api-js` `Action` instead of internal Action


gagdiez:
Currently, the wallet selector defined its own type of `Action` (https://github.com/near/wallet-selector/blob/main/packages/core/src/lib/wallet/transactions.types.ts#L70)

`near-api-js` has historically defined `Action` in a different way (https://github.com/near/near-api-js/blob/master/packages/transactions/src/actions.ts#L168)

We want the `wallet selector` to become a `Signer` of `near-api-js`, so people can simply write `naj` code and choose if they want to use `wallet-selector` or `chainsig.js` or their local wallet to sign transactions.

To achieve this, we need to remove all the internal `Action` and `Transaction` definitions, and swap them for `near-api-js` ones. This means that `signAndSendTransaction`, `signAndSendTransactions` and `signDelegate` interfaces need to change.

PR #1409, opened 2025-09-01
feat: change the wallet interface to use near-api-js Action instead of internal Action

ahnafalfariza:
## Pre-flight checklist

- [ ] I have read the [Contributing Guidelines on pull requests](https://github.com/near/wallet-selector/blob/main/CONTRIBUTING.md).
- [ ] Commit messages follow the [conventional commits](https://www.conventionalcommits.org/) spec
- [ ] **If this is a code change**: I have written unit tests.
- [ ] **If this changes code in a published package**: I have run `pnpm changeset` to create a `changeset` JSON document appropriate for this change.
- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

Migrate the wallet interface to use near-api-js Action instead of internal Action
- Add bidirectional conversion functions (`najActionToInternal`, `internalActionToNaj`)
- Update all wallet implementations to use `near-api-js` action types
- Add comprehensive transform action utilities
- Update documentation and examples

### BREAKING CHANGE: Action Type Migration
The Action type in `signAndSendTransaction`, `signAndSendTransactions`, `createSignedTransaction` now expects NAJ (NEAR API JS) actions instead of internal wallet selector action format.

#### **Before** (Old Internal Format)
```typescript
// Old internal action format
const transactions = [{
  receiverId: "contract.near", 
  actions: [
    {
      type: "FunctionCall",
      params: {
        methodName: "addMessage",
        args: { text: "Hello World" },
        gas: "30000000000000",
        deposit: "1000000000000000000000000"  // 1 NEAR
      }
    }
  ]
}];

await signAndSendTransactions({ transactions });
```

#### **After** (New NAJ Format)
```typescript
import { functionCall } from '@near-js/transactions';

// New NAJ action format using action creators
const transactions = [{
  receiverId: "contract.near",
  actions: [
    functionCall(
      "addMessage",
      { text: "Hello World" },
      BigInt("30000000000000"),
      BigInt("1000000000000000000000000")
    )
  ]
}];

await signAndSendTransactions({ transactions });
```

### **Quick Migration Tool**

If you have existing internal actions, you can use the conversion utility:

```typescript
import { internalActionToNaj } from '@near-wallet-selector/core';

// Convert old internal actions to NAJ format
const najActions = oldInternalActions.map(internalActionToNaj);

const transactions = [{
  signerId: "user.near", 
  receiverId: "contract.near",
  actions: najActions  // Now in NAJ format
}];
```

## Related issues/PRs
#1400


ahnafalfariza:
@gagdiez  I’ve migrated the interface actions to use NAJAction. Since some wallets do not yet support it, I converted them into internal actions within each wallet package. On the dapps side, they are now using Action from `near-api-js`. 

this change will probably require a major version bump due to the interface update. I think we could also take this opportunity to upgrade the `near-api-js` version alongside

does this approach make sense to you, or do you think we should handle it differently?

cc: @roman009

gagdiez:
Indeed! I think we should further take the opportunity to actually reduce dependencies on `naj` as much as possible

For example, instead of creating an internal `JSON RPC Provider`, we should ask people to provide it

gagdiez:
@ahnafalfariza indeed, that's a good idea

